Issue #31, opened 2014-03-12
can't modify immutable index when precompiling in 1.0.0


jtmalinowski:
During `rake assets:precompile`, an error appears as shown in the stack trace below.
It probably is related to changes in initializer between `0.9.0` and `1.0.0`
Stack trace is from `3.2`, but it is related to `4.0` too. 

```
rake aborted!
can't modify immutable index
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/sprockets-2.2.2/lib/sprockets/index.rb:80:in `expire_index!'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/sprockets-2.2.2/lib/sprockets/trail.rb:31:in `prepend_path'
/Users/jakub/code/react-rails/lib/react/rails/railtie.rb:29:in `block in <class:Railtie>'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/activesupport-3.2.17/lib/active_support/lazy_load_hooks.rb:34:in `call'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/activesupport-3.2.17/lib/active_support/lazy_load_hooks.rb:34:in `execute_hook'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/activesupport-3.2.17/lib/active_support/lazy_load_hooks.rb:43:in `block in run_load_hooks'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/activesupport-3.2.17/lib/active_support/lazy_load_hooks.rb:42:in `each'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/activesupport-3.2.17/lib/active_support/lazy_load_hooks.rb:42:in `run_load_hooks'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/railties-3.2.17/lib/rails/application/finisher.rb:59:in `block in <module:Finisher>'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/railties-3.2.17/lib/rails/initializable.rb:30:in `instance_exec'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/railties-3.2.17/lib/rails/initializable.rb:30:in `run'
/Users/jakub/.rvm/gems/ruby-1.9.3-p484/gems/railties-3.2.17/lib/rails/initializable.rb:55:in `block in run_initializers'
[...]

Issue #11, opened 2013-09-15
react.js not available when config.react.variant is not set


jtmalinowski:
@zpao I thought we were going to go with environment based variant, if `config.react.variant` is not explicitly set. However

```
if variant = app.config.react.variant || ::Rails.env.test?
```

makes it very clear that it needs to be configured. I think if it is as intended, then we should emphasize it a bit more in the readme. Because that bit even me a few times...

PR #36, opened 2014-03-31
Fix asset path munging to work in production env.

johnthethird:
Fix issue #31 by changing the initializer method. If you use `after_initialize` you can't then change the Sprockets path in production mode, as it becomes immutable.

Also, clean up the way the variant configs are initialized to fix issue #11.

I pulled out these specific changes from my server-rendering branch, so that they do not depend at all on server-rendering. I will rebase the server rendering changes on top of this so that they can be pulled in separately at the appropriate time.

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -5,39 +5,42 @@ module React
     class Railtie < ::Rails::Railtie
       config.react = ActiveSupport::OrderedOptions.new
 
+      # Sensible defaults. Can be overridden in application.rb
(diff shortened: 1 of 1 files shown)


vipulnsward:
@johnthethird this seems to be causing https://github.com/reactjs/react-rails/issues/148 sporadically.
Can you point me where you  picked up `after => "sprockets.environment" ` config? 
I am not able to track it down, to solve this.

rmosolgo:
I've also asked here: https://github.com/rails/sprockets-rails/issues/227

PR #40, opened 2014-04-11
Initialize railtie for all groups

sickill:
Without `:group => :all` you can't `rake assets:precompile` when you
have `config.initialize_on_precompile = false`.

I'm not sure if there are any downsides for doing this but it seems to be working fine for us.


rmosolgo:
I hope these problems are resolved with #187, please reopen if this is still an issue!

PR #59, opened 2014-07-05
Fix server-side rendering in development mode

xionon:
The renderer was using a memoized class variable to cache the
concatinated react.js and components.js for server-side rendering. This
meant that when you updated a component, the server-rendered version would
not be refreshed.

Since #setup! is called every time a watched file is updated, it seems
like the natural place to reset the memozied string.


XyuWang:
I still get the problem when add a component, but I find a way to resolve this. 
 by adding  ` config.watchable_dirs.merge!({"#{Rails.root}/app/assets/javascripts/**/" => ['jsx']})` to development.rb

PR #83, opened 2014-08-30
Minimal doc about running tests for the gem

olance:
Not much, but the info is missing and could avoid errors when trying to run tests for the first time ^^


rmosolgo:
Sorry for leaving so long, for a while there's been a short instruction on how to run the tests  in the Readme: 

image

Feel free to rebase & open another PR if you think that documentation should be extended!

PR #84, opened 2014-08-30
Remove need for global workaround and make Browserify happy

olance:
Define `self` and `window` global objects in the combined JS code so that:
- components in Coffee files can really use `window` instead of using `global` to become accessible, thus removing the need for the `var self, window, global = global || window || self;` trick in `components.js` to have the code work both on server & clients
- components/files packaged as CommonJS modules by browserify (using browserify-rails for instance) can successfully use `global` if they want to
  
  For this second point, this fix is necessary because of the self-calling function browserify uses to wrap code that accesses the `global` variable:
  
  ``` ruby
  (function (global){
     global.MyComponent = React.createClass({ ... });
  }).call(this,typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {})
  ```
  
  If `window` was used instead of `global`, it would still need the fix when compiling on the server, as `window` would not be defined otherwise.

I have not included tests, as I am not sure how to test this... 

Maybe what could serve for testing would just be to remove the `var self, window, global = global || window || self;` line from `components.js` in the dummy app, change `global.Todo = Todo` to `window.Todo = Todo` in `app/assets/javascripts/components/Todo.js.jsx.coffee` and be satisfied that tests are still successful?

Open to suggestions to complete the PR :)

### DIFF
--- a/test/dummy/app/assets/javascripts/components.js
+++ b/test/dummy/app/assets/javascripts/components.js
@@ -1,8 +1,2 @@
 //= require_self
 //= require_tree ./components
-
-// This is because we compile this file into a JS VM
-// for server rendering, and some components may be
-// .coffee and wrapped in a func, so they need a
-// global to glom on to.
-var self, window, global = global || window || self;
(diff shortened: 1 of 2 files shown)


olance:
@zpao, has @jwietelmann's objection been taken into account?

PR #89, opened 2014-09-12
disable the Turbolinks cache, fixes #87

### LINKED ISSUE #87: Components are not unmounted when navigating back
Components on the current page are not unmounted when navigating to a page cached by Turbolinks (e.g. back button).

Events for an initial page load:

1. `page:change` *<-- components are mounted*

Events for a normal page visit (e.g. clicking a link):

1. `page:before-change` *<-- components are unmounted here...*
2. `page:fetch`
3. `page:receive` *<-- ...or here with #85*
4. `page:change` *<-- components are mounted*
5. `page:load`

Events for a cached page (e.g. back button):

1. `page:change` *<-- components are mounted*
2. `page:restore`

As you can see, **neither of the unmount events are fired for cached pages**.

My initial though was to unmount on `page:change` and mount on both `page:load` and `page:restore`, but this obviously won't work for the initial page load.

The only solution I can currently think of is to turn off Turbolinks caching via `Turbolinks.pagesCached(0)`

meleyal:


rmosolgo:
@meleyal  I'm interested in getting this in, but it doesn't merge cleanly since redoing the UJS. Are you interested in updating?

meleyal:
Will do

meleyal:
Updated

rmosolgo:
Thanks!

image

:tada:

PR #91, opened 2014-09-14
Expose mount/unmount and allow them to be scoped by a selector.

mschulkind:
Looking for comments on this before I attempt to add tests.

I'm currently using this for an app which has only a few react components, including some that are not present on page load and may disappear later, so I need to mount/unmount components from parts of the page manually in other javascript.


rmosolgo:
@mschulkind It'd be great to get this in for 1.0, are you interested in rebasing & updating etc?

rmosolgo:
Thanks for the inspiration, this feature was implemented in #221 !

Issue #42, opened 2014-05-14
Difficult to debug non-browser context class of JS errors


lencioni:
When using the prerender option, react-rails will rescue any JavaScript errors. This will log it and continue, which will just render the react ujs tag, and when the browser tries to render the component it will most likely encounter the same error and throw to the browser console for a better debugging experience.

While this may be the case for some JS errors, it is not the case for errors that happen when running JS not in the browser context but that don't happen when running JS in the browser context. For example, the `window` object is available in the browser context but not available in the non-browser context, causing errors like "window is not defined". By rescuing these exceptions, seeing that something is broken in these situations is difficult.

What do you think about not rescuing JS exceptions when prerendering? This would make debugging in these situations quite a bit less mysterious. Folks who want a better debugging environment for JS errors that happen on the client side should be able to easily disable prerender.

PR #92, opened 2014-09-15
Raise any JS error when prerendering

rmosolgo:
A simple attempt at solving #42. Even with better_errors it's not pretty:

image

But I'm not sure what else can be done.  Have a look at how `ProgramError` is raised: https://github.com/sstephenson/execjs/search?utf8=%E2%9C%93&q=ProgramError

It looks like it only comes with a `message` -- no stack trace. Any ideas on how to make this more useful?

### DIFF
--- a/lib/react/renderer.rb
+++ b/lib/react/renderer.rb
@@ -3,6 +3,13 @@ require 'connection_pool'
 module React
   class Renderer
 
+    class PrerenderError < RuntimeError
+      def initialize(component_name, props, js_message)
+        message = "Encountered error \"#{js_message}\" when prerendering #{component_name} with #{props.to_json}"
+        super(message)
+      end
+    end
+
     cattr_accessor :pool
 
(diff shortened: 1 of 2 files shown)


lib/react/renderer.rb
```diff
@@ -56,13 +63,8 @@ def render(component, args={})
         }()
       JS
       context.eval(jscode).html_safe
-    # What should be done here? If we are server rendering, and encounter an error in the JS code,
-    # then log it and continue, which will just render the react ujs tag, and when the browser tries
-    # to render the component it will most likely encounter the same error and throw to the browser
-    # console for a better debugging experience.
     rescue ExecJS::ProgramError => e
-      ::Rails.logger.error "[React::Renderer] #{e.message}"
+      raise PrerenderError.new(component, args, e)
```
trungpham: can we log the stack trace too?

PR #93, opened 2014-09-21
Only use turbolinks events if supported

felixbuenemann:
The current code always uses Turbolinks events if it is loaded, but this
doesn't mean it is supported by the browser. This change checks if
turbolinks is supported and otherwise uses the native event handlers.

Without this fix ujs components are not rendered in IE8.


felixbuenemann:
Thanks, @rmosolgo!

PR #98, opened 2014-10-17
Reload all dependent JS when files change

tonymarklove:
Use a custom watcher/reloader middleware to find all the dependent
files of component_filenames and watch them for changes.

The reloader sits outside of the Rails stack, but makes use of
ActiveSupport::FileUpdateChecker.

This has a few benefits:

1. Non-jsx files can easily be watched. For example, changes to
   components.js and other helper files are detected.

2. Newly required files can be detected and added to the list of
   watched files for future updates.

3. We don’t force Rails to reload its own Ruby classes when only .js
   or .jsx files have changed.

One potentially controversial change, which didn’t seem to provide
much benefit before:

- Moved the concatenated source files for components.js and react.js
  out of the app config. They are now encapsulated in the
  after_initialize block.

### DIFF
--- a/lib/react/rails.rb
+++ b/lib/react/rails.rb
@@ -1,3 +1,4 @@
 require 'react/rails/railtie'
 require 'react/rails/engine'
+require 'react/rails/reloader'
 require 'react/rails/view_helper'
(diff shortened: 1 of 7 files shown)


rmosolgo:
I think this is handled by the gem now, because 

- We watch `.jsx` files ([src](https://github.com/reactjs/react-rails/blob/master/lib/react/rails/railtie.rb#L24))
- To reload, we clear out existing renderers ([src](https://github.com/reactjs/react-rails/blob/master/lib/react/rails/railtie.rb#L88))

If that's not sufficient or if I've missed something, please feel free to re-open & rebase on master!

PR #101, opened 2014-10-22
Adds benchmark option; Adds more tests;

pboling:
• Adds benchmark option
• Adds more tests


pboling:
Building a better boat here: https://github.com/pboling/react-rails-benchmark_renderer

rmosolgo:
Very cool! Glad to see those refactors are paying  off. 

Yes, I want to get a new version out soon, but we're blocked behind https://github.com/rails/sprockets-rails/issues/257 (And I don't have the nerve to dive in and figure it out!)

PR #102, opened 2014-10-22
Add support for saving references to components

arnemart:
When rendering a component using the react_component view helper,
pass `reference: 'myReferenceName'` in the options hash to make
react_ujs store a reference to the rendered component at
`window.myReferenceName`. This is useful for scripts that need to
make use of a component after a page has loaded, without needing
to re-render an extra time.


rmosolgo:
sounds like this could be accomplished in a more explict, context-specific way

PR #115, opened 2014-11-17
Add suffix code for context during server rendering

azhuravlev:
Allow to send and execute js code during context compiling, before component eval.
This can be usefull, then components are using some variables generated on server, such as generated by Gon https://github.com/gazay/gon


rmosolgo:
Hoping to support more customization of the view helper with #346 !

That way you could distribute a custom helper & renderer in a gem and hook them up with a railtie.

rmosolgo:
I hope that new extension point is working for you!

Issue #116, opened 2014-11-18
Proposal: Return UJS as a public API


robrobbins:
Doing a SPA with react I have found it beneficial to have `unmountReactComponents` publicly accessible. I have modded the `react_ujs` file in this project to do this easily enough. I think there is an interesting discussion that can take place around this:

1) making it a 'revealing module pattern' is simple but where should it live really? A `ReactUjs` obj ? I actually just append it to React itself as `React.ujs` accessing them, obviously, as `React.ujs.unmountComponent` etc...

2) I realize this may be outside of the scope you want to address with the gem. That being the case I can fork and maintain it separately

3) If interested I can work up a PR with the appropriate tests

rmosolgo:
closed because #117 was merged :confetti_ball:

PR #117, opened 2014-11-19
Expose the React mounting and un-mounting mechanisms

robrobbins:
Re: #116 

This is my initial commit, all within a `react_ujs_p.js` file so that we may discuss it further. I'll get to the tests tomorrow, though I expect the existing to pass. I do need to write a few new ones.

Currently I am using this modded version is a non-turbolinks SPA and it works fine.


rmosolgo:
@zpao any objections to going forward like that ^^ ?

@robrobbins  He said he'd be busy so if we don't hear from him in 48 hours, I think we should take that as agreement :)

robrobbins:
sure. i'll get on it later today then

zpao:
:+1: sounds good! Sorry for the delay, this project is unfortunately low on the priority list with everything else :(

robrobbins:
Done

rmosolgo:
:+1: you're a good man @robrobbins 

dropped this in locally just to make double-sure & looks good to me

robrobbins:
:+1: Thanks

Did the same, been using it on the redesign of a big commercial app. Doing fine.

:shipit:

rmosolgo:
:money_with_wings:

jmagoon:
This is a great addition. Quick question, is there any reason to leave the JSON inside of the data-react-props attribute after you've called React.render()? My current application renders about 5000 objects to the page, and having those sit in the attribute causes the developer console to crawl and sometimes crash.

My current approach is just to call node.removeAttribute(PROPS_ATTR) right after the render step in the MountComponents function. Is there a reason to keep that attribute around?

rmosolgo:
@jmagoon I'd be in favor of that change. 

The only downside I can think of is that components wouldn't be re-mountable. But, is that something that  matters? Does anyone depend on that?

Actually, I think I do that sometimes: arbitrarily fire `page:change` (faking the Turbolinks event) in order to mount components in a just-loaded modal. At present, it just doesn't matter because it re-mounts with the same `data-react-class` and `data-react-props` (and React doesn't end up doing anything to the DOM). So maybe if you removed `data-react-props`, you could also remove `data-react-class`, so that calling `ReactRailsUJS.mountComponents()` would just miss those things all together.

Oh wait, that wouldn't work because `unmountComponents` depends on that `data-react-class` attribute. :S

Anyways, I'm open to something like that ^^, especially if we could get something like #91 worked in. (Then I could stop my stupid hack for mounting new components!)

robrobbins:
i can look into it

rmosolgo:
@jmagoon Turns out we need to keep those props in the DOM to support turbolinks, see #159

PR #121, opened 2014-11-23
Adding example about how to disabling automatic unmounting component

andreynering:
Adding an example for who don't want components being unmounted.


dapi:
Thanks for a tip )

rmosolgo:
For non-Turbolinks settings, this has been disabled by default (since https://github.com/reactjs/react-rails/pull/318)

PR #132, opened 2014-12-11
Saving react source to Dir.tmpdir

golmansax:
We needed to make this change because our CI (Tddium/Solano) runs workers in parallel and `tmp/` inside the repo is not safe.  Thoughts about this?  Thanks!


golmansax:
Wow sorry, I totally forgot about this pull.  I'll close this out and switch to using the `react-rails` gem, and reopen the pull if we run into this issue again.

PR #136, opened 2014-12-23
feat(React::JSX): add JSX.transform_options

rmosolgo:
I want the feature from #107  :)

Now you can say in `application.rb`:

``` ruby
config.react.jsx_transform_options = {harmony: true}
```

And you'll get es6 features in your `.jsx` assets. What do you think about this implementation?
- Take a config
- Assign it to a module variable 
- Use that when rendering with `React::JSX::Template`

I had to change the requires in the dummy `application.js` so that the ES6 file wouldn't wreck other tests.

I've used it with a test component in my app and it works!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,10 +68,13 @@ Component = React.createClass
     `<ExampleComponent videos={this.props.videos} />`
 ```
 
-You can use the `--harmony` option by adding a configuration to `application.rb`:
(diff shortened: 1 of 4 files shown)


rmosolgo:
Oops, had to add a leading `/` for the tests to pass

PR #138, opened 2015-01-02
Fix invalid filename in README.

technion:
The file referenced in the instructions does not exist. I believe this change will reflect the intended file.


rmosolgo:
:+1: thanks, good catch!

PR #139, opened 2015-01-02
Upgrade react-source to 0.12.2

jordanstephens:
Upgrading react-source to 0.12.2

Tests were failing when expecting a single quoted string in transformed JSX. It looks like this can be traced back to facebook/react@eddbb0c, so I'm updating the expectation to a double-quoted string—see 71a01bc.

### DIFF
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'execjs'
   s.add_dependency 'rails', '>= 3.1'
-  s.add_dependency 'react-source', '0.12'
+  s.add_dependency 'react-source', '~> 0.12'
   s.add_dependency 'connection_pool'
 
   s.files = Dir[
--- a/test/jsxtransform_test.rb
+++ b/test/jsxtransform_test.rb
@@ -10,7 +10,7 @@ EXPECTED_JS_2 = <<eos
 (function() {
   var Component;
 
-  Component = React.createClass({displayName: 'Component',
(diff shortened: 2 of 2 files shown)


rmosolgo:
I thought leaving out the minor version was be design -- it installs the latest minor version by default, right?

zpao:
`gem install react-rails` uses the 0.12.x.y release, which does specify patch versions. Only the stable releases have been published. `master` however is more fluid and uses the latest instead of locking because we don't really have version numbers there. If you did this against the 0.12 branch (https://github.com/reactjs/react-rails/tree/0.12) then I would go ahead and merge this in but we probably shouldn't do this for master.

Want to do this against the other branch and then I'll publish the v0.12.2.0 gem.

jordanstephens:
Thanks @zpao, I'll open a new PR on the 0.12 branch.

Before I close this, I'm still a little bit confused:

> master however is more fluid and uses the latest instead of locking 

I believe the current dependency definition on master *is* locked (on 0.12.0):

`s.add_dependency 'react-source', '0.12'`

I just double checked the version in my Gemfile.lock after bundling from master and I am indeed seeing `react-source (0.12.0)`.

Don't we want to use the following on master instead?

`s.add_dependency 'react-source', '~> 0.12'`

Thanks again!

zpao:
Yea, I was actually going to ask. I think you're right that we want the `~>`. I think that's safe to go ahead and do.

jordanstephens:
Just rebased this branch to include the `~>`.

Thanks @zpao!

rmosolgo:
Nice!

PR #140, opened 2015-01-05
Upgrade react-source to 0.12.2

jordanstephens:
Follow up to #139, using 0.12 as a target branch instead of master.

See previous discussion at #139.


jordanstephens:
Had a little hiccup there in the tests with an upgrade to actionpack 4.2. It looks like this has already been addressed on master.

zpao:
:thumbsup: Thanks!

PR #141, opened 2015-01-06
Cache bundler installs on TravisCI for faster builds

zpao:
Also removes explicit install of bundler. Travis has it already.


zpao:
Anybody have any ideas why this would be failing? I thought this would be an easy way to shave a few minutes off our build times but no dice (even after I re-added the explicit bundler install)

PR #142, opened 2015-01-06
Report test coverage to codeclimate

zpao:
I'm really not sure about where this gem dependency should be specified. I remember being confused by this for dev dependencies in the past and just shoved everything in the gemspec. But I just listened to what Code Climate told me to do to set this up. Correct me if I'm wrong. cc @rmosolgo


rmosolgo:
Yeah, I don't know that bit too well either. I bet that example is for a Rails gemfile. Here's a working example where it's added with `add_development_dependency`:

https://github.com/jfelchner/ruby-progressbar/blob/master/ruby-progressbar.gemspec

Maybe that will magically fix the JRuby failure...

rmosolgo:
see also https://github.com/goldstar/blackbeard/blob/master/blackbeard.gemspec

rmosolgo:
finished up in #274 :tada:

PR #145, opened 2015-01-09
README update

uberllama:
The README me shows unnecessary double nesting on the jbuilder example. Corrected to use standard collection syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,11 +165,9 @@ which generates JSON like this:
 This is not suitable for ReactJS props, which is expected to be a key-value object. You will need to wrap your index.json.jbuilder node with a root node, like so:
 
 ```ruby
-json.messages do |json|
-  json.array!(@messages) do |message|
-    json.extract! message, :id, :name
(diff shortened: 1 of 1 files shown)


rmosolgo:
I don't know the first thing about jbuilder, @xionon what do you think of this?

uberllama:
Would it help if I said "trust me"? :) I've used it on many projects and this is standard wrapped collection syntax. You very rarely use `array!` in Jbuilder, precisely because you typically need to support multiple nodes on a response.

rmosolgo:
Haha, I trust you, but I wouldn't be doing a good job if I didn't double check :P 

I'll give it a try later if we don't here from xionon. I figured I'd ask him since he wrote that feature to start with!

xionon:
Same output, simpler syntax, :+1: from me

rmosolgo:
:boom: Thanks guys!

uberllama:
kaplah

PR #150, opened 2015-01-20
Faster test on Travis CI

tricknotes:
before: [about 4 min](https://travis-ci.org/tricknotes/react-rails/builds/47638822)
after: [about 3 min](https://travis-ci.org/tricknotes/react-rails/builds/47639547)


zpao:
I have no idea why this passed but #141 didn't (even when I dropped the bundler install). Oh well. Thanks!

PR #153, opened 2015-01-24
Correctly pass `max_renderers` config to setup

xionon:
This has been broken for a while :grimacing: 

Basically, the code has been passing the wrong variable in, so the render pool size was always defaulting to `10`.


zpao:
Thanks!

PR #154, opened 2015-01-27
Document `config.watchable_files`

ConradIrwin:
This makes it easier to figure out how to set up code-reloading of files with a  `.cjsx` extension (I'm using https://github.com/jsdf/sprockets-coffee-react).

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,16 @@ end
 
 ```
 
+In production your compiled JS is cached by each JS VM. In development this cache is cleared whenever you change a `*.jsx*` file in `app/assets/javascripts`. In order to clear this cache when you edit files that do not fit this pattern, you need to tell Rails which paths to watch for changes.
+
+```ruby
+# config/environments/application.rb
+# These are the defaults if you dont specify any yourself
+MyApp::Application.configure do
+  config.watchable_files.concat Dir["#{app.root}/app/assets/javascripts/**/*.jsx*"]
+end
+```
+
 ## CoffeeScript
 
 It is possible to use JSX with CoffeeScript. The caveat is that you will still need to include the docblock. Since CoffeeScript doesn't allow `/* */` style comments, we need to do something a little different. We also need to embed JSX inside backticks so CoffeeScript ignores the syntax it doesn't understand. Here's an example:
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ In production your compiled JS is cached by each JS VM. In development this cach
 # config/environments/application.rb
 # These are the defaults if you dont specify any yourself
 MyApp::Application.configure do
-  config.watchable_files.concat Dir["#{app.root}/app/assets/javascripts/**/*.jsx*"]
+  config.watchable_dirs[Dir["#{app.root}/app/assets/javascripts"]] = ["*.jsx*"]
 end
 ```
 


README.md
```diff
@@ -317,6 +317,16 @@ end
 
 ```
 
+In production your compiled JS is cached by each JS VM. In development this cache is cleared whenever you change a `*.jsx*` file in `app/assets/javascripts`. In order to clear this cache when you edit files that do not fit this pattern, you need to tell Rails which paths to watch for changes.
+
+```ruby
+# config/environments/application.rb
+# These are the defaults if you dont specify any yourself
+MyApp::Application.configure do
+  config.watchable_files.concat Dir["#{app.root}/app/assets/javascripts/**/*.jsx*"]
```
xionon: AFAICT this will only watch files that exist when you boot your development server, new files added after that will not be watched. Wouldn't it make more sense to suggest [watchable_dirs](http://api.rubyonrails.org/classes/Rails/Railtie/Configuration.html#method-i-watchable_dirs) here?

ConradIrwin: Good point, fixed! Interestingly enough this approach seemed to work for me because I used Dir["#{Rails.root}/app/assets/javascripts/react/**/*"]), which contained the directories too.

xionon:
So, I finally got around to testing this, and I'm not sure that setting `watchable_dirs` is actually needed.

With a fresh Rails app, and both `react-rails` (master) and `sprockets-coffee-react` installed, I have no trouble in development mode getting cjsx files to auto-reload. New files and changes are both picked up on reload.

Are you putting your cjsx files in `app/assets/javascripts/components`?

ConradIrwin:
I'm not. They're in `app/assets/javascripts/react/components`.

Conrad

xionon:
That works by default for me, too...

The railtie already does basically the same thing that you add -- https://github.com/reactjs/react-rails/blob/master/lib/react/rails/railtie.rb#L19-L21

What version of Rails are you using? I'm thinking it's either a bug with our integration with previous versions of Rails, or a bug in your implementation.

ConradIrwin:
I'm using `.cjsx` files for coffee-script. We could change the gem to watch those too.

xionon:
Ahhh, I see, it's a prerender issue. I misunderstood, sorry!

I'd rather not go down the path of "documenting Rails features" if we can help it. I think I would rather see a change to the railtie in the gem, maybe to watch `.*jsx` instead of `.jsx` - would that help?

ConradIrwin:
That sounds great, thanks!

Conrad

mpereira:
@xionon people using side builds with browserify or webpack might not be using files with a ".jsx" extension, but ".js" instead (my case). I had to add `config.watchable_files.concat(['app/assets/javascripts/bundle.js'])` to config/development.rb to get the JS VM reloading the JS bundle when it changed, for prerender.

dfguo:
@mpereira I'm in a similar situation. I'm using webpack, but putting 
```
config.watchable_files.concat(['app/assets/javascripts/bundle.js']) 
```

to config/development.rb doesn't work for me...

dfguo:
works after i used absolute path:

config.watchable_files.concat Dir["#{Rails.application.root}/app/assets/javascripts/bundle.js"]

rmosolgo:
Sorry, I think this has been cold from even before I started maintaining this. 

Please feel free to reopen if you think this is still an issue!

rmosolgo:
Using FileUpdateChecker in 2.0 will be more configurable and also catch newly-added files: https://github.com/reactjs/react-rails/pull/682

chrisjingram:
If anyone lands here from Google and is still having an issue with their files not reloading in SSR, my issue was caused by react-rails not seeing the changes to .tsx files (as I'm using Typescript) and therefore not reloading the renderers.

You can tell react-rails to watch for .tsx files by adding this to your `config/development.rb`:

```ruby
config.react.server_renderer_extensions = ["jsx", "js", "tsx", "ts"]
```

PR #163, opened 2015-02-03
Use direct eval in react_ujs component mounting fallback

### LINKED ISSUE #162: Mounting of components using indirect eval call has cross-browser issues
In `react_ujs`, in the `mountComponent()` function, `react-rails` is using an [indirect `eval` call as a fallback to element loading](https://github.com/reactjs/react-rails/blob/07dcbcb/lib/assets/javascripts/react_ujs.js#L34). This approach has cross-browser issues, as ES3 allows browsers to throw on indirect eval calls. It will also leak all the variables to global scope, which might or might not be a good idea.

What indirect eval, you might say? I suggest you read the following blog post in order to fully understand the quirks and weirdosities of eval.

http://perfectionkills.com/global-eval-what-are-the-options/

Especially regarding the indirect eval and cross-browser issues, check this part of the blog post — [Indirect eval call. In practice.](http://perfectionkills.com/global-eval-what-are-the-options/#indirect_eval_call_in_practice)

Maybe using plain, direct `eval` call would be better, as it would evaluate the code in the local context of `mountComponent` function, have much better consistency between different browsers and would not leak variables in to the global context. What do you think? What I'm proposing is simply changing the fallback mechanism to just use `eval(className)` instead of `eval.call(window, className)`.

EDIT: Actually made a PR for the proposed change: #163

valscion:
Closes #162 

http://perfectionkills.com/global-eval-what-are-the-options/

### DIFF
--- a/lib/assets/javascripts/react_ujs.js
+++ b/lib/assets/javascripts/react_ujs.js
@@ -31,7 +31,7 @@
         
         // Assume className is simple and can be found at top-level (window).
         // Fallback to eval to handle cases like 'My.React.ComponentName'.
-        var constructor = window[className] || eval.call(window, className);
+        var constructor = window[className] || eval(className);
         var propsJson = node.getAttribute(window.ReactRailsUJS.PROPS_ATTR);
         var props = propsJson && JSON.parse(propsJson);
         
--- a/lib/assets/javascripts/react_ujs.js
+++ b/lib/assets/javascripts/react_ujs.js
@@ -31,7 +31,7 @@
         
         // Assume className is simple and can be found at top-level (window).
         // Fallback to eval to handle cases like 'My.React.ComponentName'.
-        var constructor = window[className] || eval(className);
+        var constructor = window[className] || eval('window.' + className);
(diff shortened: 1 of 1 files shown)


valscion:
Huh, I have no idea why the tests fail in CircleCI. They pass on my machine with the same appraisal command ran in Travis

jakubmal:
I rerun tests for your PR - probably a timing issue. Still, this would need a rebase to be merged.
However, I can't see any repro case, where this ```eval``` wouldn't work as expected. We assume that react classes are either global or referenced as fields of global objects.

valscion:
> However, I can't see any repro case, where this `eval` wouldn't work as expected. We assume that react classes are either global or referenced as fields of global objects.

Yes, I actually pondered on this myself for some time ago. I think I have turned my head around the subject and come to the conclusion that this indirect eval actually works good enough and browsers which don't support indirect eval are not even the browsers React should work on (I believe).

So maybe this can be closed, as well as the issue related to this. :confused:

jakubmal:
Alright! Thanks for this PR.

PR #166, opened 2015-02-05
feat(ujs) Pass contents of content tag to React component.

danott:
### Usage

```erb
<%= react_component('FilterChildren') do %>
  <div>Some child HTML that you want to use in your component</div>
  <div>Even though we'll likely be using React's dangerouslySetInnerHTML property, it's better than writing jQuery and tracking down the magic classes and markup</div>
<% end %>
```

```jsx
var FilterChildren = React.createClass({
  propTypes: {
    __html: React.PropTypes.string
  },

  someTransformationOfRawHTML: function() {
    /* do something with this.props.__html */
  }, 

  render: function() {
    return <div dangerouslySetInnerHTML={{__html: this.someTransformationOfRawHTML()}} />;
  }
});
```

In your component, `this.props.__html` will be available. The value is the string `domNode.innerHTML`, where `domNode` is the node being mounted on.

### DIFF
--- a/lib/react/rails/view_helper.rb
+++ b/lib/react/rails/view_helper.rb
@@ -7,6 +7,7 @@ module React
       # on the client.
       #
       def react_component(name, args = {}, options = {}, &block)
+        args[:__html] = capture(&block) if block.present? && options[:__html]
         options = {:tag => options} if options.is_a?(Symbol)
         block = Proc.new{concat React::Renderer.render(name, args)} if options[:prerender]
 
@@ -16,13 +17,12 @@ module React
           data[:react_props] = React::Renderer.react_props(args) unless args.empty?
         end
         html_tag = html_options[:tag] || :div
-        
+
         # remove internally used properties so they aren't rendered to DOM
         html_options.except!(:tag, :prerender)
-        
+
         content_tag(html_tag, '', html_options, &block)
       end
-
     end
   end
 end


xionon:
I don't understand why you wouldn't just put this in the component itself?
It seems like splitting the HTML template up into two locations like this
would (a) break isomorphism, and (b) make things hard to maintain.

If you're trying to get some server-generated properties, e.g. i18n
translations, into your component, it seems like it would be better to do
this via props that you pass in.

danott:
@xionon [an example](http://jsfiddle.net/danott/0c0jyd09/1/) of the _concept_.

In Rails, this might look like

```erb
<% @records = SomeModel.all %>
<%= react_component('RevealMoreChildren', { begin: 4 }) do %>
  <%= render @records #using partials %>
<% end %>

<% @other_records = SomeOtherModel.all %>
<%= react_component('RevealMoreChildren', { begin: 2 }) do %>
  <%= render @other_records #using partials %>
<% end %>
```

The use case is that I don't want to write a custom component for every list of items that might be wrapped by `RevealMoreChildren`. As noted in the PR, having react handle this rather than magical class names and markup is a win over jQuery soup, but it's a loss if we are writing a component for every use case of a simple behavior.

xionon:
Why not just use render and pass it in as a property?

```
<% @records = SomeModel.all %>
<%= react_component( 'RevealMoreChildren', { __html: render(@records) }) %>
```

danott:
I like the explicitness . That's definitely a workable solution for rendering another partial.

Trying to think about how to pass in the param for a non-partial content tag.

```erb
<% html_prop = capture do %>
  <h1>Arbitrary markup</h1>
  <p>It could be huge</p>
<% end %>
<%= react_component( 'RevealMoreChildren', { __html: html_prop }) %>
```

That works, but doesn't read as nice.

rmosolgo:
I agree with @xionon -- passing in strings of HTML is not a common React practice, so I don't think react-rails needs to go out of its way to support it.

danott:
I've come around to thinking this is not a good idea to include in the main helper. I've created a helper in my app that does what I need.

nicolas-besnard:
@danott can you share your helper ?

danott:
@nicolas-besnard Absolutely! It's a thin wrapper around `react_component`

```ruby
# app/helpers/react_helper.rb
module ReactHelper
  def react_component_with_content(name, args = {}, options = {}, &block)
    args[:__html] = capture(&block) if block.present?
    react_component(name, args, options, &block)
  end
end
```

nicolas-besnard:
I love it ! Thanks

borisrorsvort:
I guess you'll still have to pay attention while doing that since: https://facebook.github.io/react/tips/dangerously-set-inner-html.html

PR #167, opened 2015-02-06
Try to fix specs

rmosolgo:
I have no idea if this will work

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,8 +13,6 @@ Rake::TestTask.new(:test) do |t|
   t.libs << 'lib'
   t.libs << 'test'
   t.pattern = ENV['TEST_PATTERN'] || 'test/**/*_test.rb'
-  t.verbose = ENV['TEST_VERBOSE'] == '1'
-  t.warning = true
 end
 
 task default: :test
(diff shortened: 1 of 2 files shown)


rmosolgo:
WHY WHY WHY 

image

image

Rakefile
```diff
@@ -13,8 +13,6 @@ Rake::TestTask.new(:test) do |t|
   t.libs << 'lib'
   t.libs << 'test'
   t.pattern = ENV['TEST_PATTERN'] || 'test/**/*_test.rb'
-  t.verbose = ENV['TEST_VERBOSE'] == '1'
-  t.warning = true
```
vipulnsward: I think this is a bad idea. Warnings are good to come across things we need to fix

rmosolgo: Alright, I'm happy to leave those in. It just seems like the Travis logs are full of warnings from gems, for example: https://travis-ci.org/reactjs/react-rails/jobs/49538440#L312

xionon: Why not allow a silence flag, instead?

```ruby
t.verbose = ENV['SILENCE_WARNINGS'] != '1'
t.warning = ENV['SILENCE_WARNINGS'] != '1'
```

vipulnsward: After checking this again, the warnings are from rails 3.2, etc. Those versions are no longer maintained and won't be patched. So :+1: for this change right now.

vipulnsward:
@rmosolgo I investigated this failure, and looks like fixing  'coffee-script-source' version takes care of this issue.
```ruby
s.add_dependency 'coffee-script-source', '1.8.0'
```
Need to check why this behaviour.

rmosolgo:
ohhh i see ... https://github.com/jashkenas/coffeescript/issues/3829

PR #168, opened 2015-02-06
chore(gems) use coffee-script-source 1.8.0

rmosolgo:
Looks like some bug in `coffee-script-source 1.9.0` 

https://github.com/jashkenas/coffeescript/issues/3829

curious if this will do the trick


vipulnsward:
Fixed :+1: 
We could get rid of warnings too, since some of the versions we are testing are unmaintained.

rmosolgo:
:tophat:s  off to @vipulnsward !

PR #170, opened 2015-02-10
Update readme for bundler 1.8 API change

nacengineer:
Bundler 1.8 is stricter regarding declaring the source of your gems, since this gem is being 
sourced from github you need to explicitly declare it in Gemfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,11 @@ As with all gem dependencies, we strongly recommend adding `react-rails` to your
 # If you missed a warning at the top of this README - this is still in development
 # which means this version is not pushed to rubygems.org
 
+# bundler <  1.8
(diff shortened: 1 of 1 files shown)


nacengineer:
also I had to --vendor my gems to run a bundle exec rails console. That may just be on my end.

xionon:
I don't think http://gems.github.com exists anymore, and [as of 1.8, the :github key appears to still be supported](https://github.com/bundler/bundler/blob/v1.8.0/man/gemfile.5.ronn#L341-L359).

Were you doing this to solve a specific problem? Can you provide a repro case?

nacengineer:
well that part was required with the 1.8.0 bunder on my machine. however I think it was a bug and not an API change. It looks like they addressed this since then

https://github.com/bundler/bundler/blob/master/CHANGELOG.md

As such this PR can be closed.

PR #172, opened 2015-02-12
Allow greater coffee-script-source versions.

gunn:
I assume their's no reason we can't use version 1.9.0.


rmosolgo:
It caused random failing specs, see https://github.com/reactjs/react-rails/pull/168

Open to other suggestions how to deal with this!

oguzbilgic:
:+1: for coffescript 1.9.0 support.

rmosolgo:
They just released `coffee-script-source 1.9.1` which seems fixed. I re-ran this one but it still installed `1.9.0`. I'm going to resolve this with #180

Issue #177, opened 2015-02-18
Readme clarification, Jbuilder


noinkling:
Under "Important Note" in the readme there is some advice about Jbuilder returning an array as the root:

> This is not suitable for ReactJS props, which is expected to be a key-value object. You will need to wrap your index.json.jbuilder node with a root node...

Being new to React, I'm not exactly sure what this refers to, as I can pass an array through as a prop just fine. Does it refer to `setState()` needing an object? If so, could the explanation be made clearer about that being the reason? Or is there some other underlying reason?

**Edit:** just realized that it's likely referring to use with the Rails view helper as outlined above it. If that's the case then it could probably be made a bit more obvious that it's more for the benefit of the helper than for React itself (even though I know it's usually seen as best practice regardless).

xionon:
You can pass an array as a prop, yes. You cannot pass an array as all the
props.

`react_component`'s second parameter is expected to be either an object
that responds to `to_json`, or a string from an object that has already
been transformed into JSON. For example, you could pass `{:foo => "Bar"}`,
or `{:foo => "Bar}.to_json`. Either way, it expects the output to be a
key-value map that React can use as props.

`render(template: "some-jbuilder-template.json.jbuilder")` just returns a
string, which can be passed into `react_component` as the second argument.
This is a nice way to render more complex JSON than could be expressed
nicely with `#to_json`, as well as reuse the JSON templates for both
pre-rendering and client-side fetching.

The note is to clarify that Rails default Jbuilder scaffolding for the
index view will have `render(template: "index.json.jbuilder")` return the
list represented as a flat array, not a map.

As an example, if you were hard-coding it, this would work:
`<%= react_component 'MyComponent', '{ posts: [ {id: 1, name: "Post 1"} ]
}' %>`

This would not work:
`<%= react_component "MyComponent", '[ {id: 1, name: "Post 1"} ]' %>`

This second example is the default format that the Rails scaffold will
render to. Note that in the second example, we are just passing an array as
the props. React won't know what to do with that.

If you can think of a way to clarify the README on this, a PR would be
greatly appreciated!

noinkling:
Thanks for explaining, I opened a PR at #178 that will hopefully dispel any ambiguity. I think the issue was mostly that it wasn't clear that "props" referred to a single collection of props, rather than props in general.

PR #178, opened 2015-02-18
Readme: clarify Jbuilder root node recommendation

noinkling:
As per #177. I think this makes things clearer.


xionon:
:+1: This is much better, thank you!

rmosolgo:
Thanks!

PR #180, opened 2015-02-20
fix(gems) add coffeescript 1.9.1

rmosolgo:
They've published coffee-script-source 1.9.1, let's see if that resolves the test failure issue

### DIFF
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'jbuilder'
 
   s.add_dependency 'execjs'
-  s.add_dependency 'coffee-script-source', '1.8.0'
+  s.add_dependency 'coffee-script-source', '~>1.9'
   s.add_dependency 'rails', '>= 3.1'
   s.add_dependency 'react-source', '~> 0.12'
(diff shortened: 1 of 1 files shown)


vipulnsward:
:clap:

rmosolgo:
Yuck, apparently it's not so simple, it made master fail: https://travis-ci.org/reactjs/react-rails/jobs/51537751

given the problems `~>1.8` is causing, i guess we should just leave `~>1.9` ??

vipulnsward:
Oh, right that will pickup 1.9, we can define hard dependency on 1.9.1 for now.
>= 1.9.1

rmosolgo:
It looks to me like that job _did_ bundle `1.9.1`, or am I misreading it?


https://travis-ci.org/reactjs/react-rails/jobs/51537751#L112
https://travis-ci.org/reactjs/react-rails/jobs/51537751#L194
https://travis-ci.org/reactjs/react-rails/jobs/51537751#L268

vipulnsward:
Its hard to track down the issue with https://travis-ci.org/reactjs/react-rails/jobs/51537751 , since its not reproducible. It passes for me for ruby 1.9.3

PR #182, opened 2015-02-22
Add the ability to choose the json engine

mickey:
This adds the ability to change the JSON backend:
```ruby
  config.react.json_engine = Oj
end
```

### DIFF
--- a/test/react_renderer_test.rb
+++ b/test/react_renderer_test.rb
@@ -1,19 +1,18 @@
 require 'test_helper'
 
 class ReactRendererTest < ActiveSupport::TestCase
-
   test 'Rendering with a different json engine' do
-    require 'oj'
-    ::Oj.default_options = { :mode => :compat }
-
     cfg = Rails.application.config.react
     React::Renderer.setup!(cfg.react_js, cfg.components_js, {
-      :size => cfg.max_renderers, :timeout => cfg.timeout, :json_engine => ::Oj
+      :size => cfg.max_renderers, :timeout => cfg.timeout, :json_engine => Class.new do
+        def self.encode(*args)
+          "{\"todos\":[\"custom1\",\"custom2\",\"custom3\"]}"
+        end
+      end
     })
 
     result = React::Renderer.render "TodoList", :todos => %w{todo1 todo2 todo3}
-    assert_match(/todo1.*todo2.*todo3/, result)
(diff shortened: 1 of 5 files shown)


test/react_renderer_test.rb
```diff
@@ -2,6 +2,24 @@
...
+    ::Oj.default_options = { :mode => :compat }
+
+    cfg = Rails.application.config.react
+    React::Renderer.setup!(cfg.react_js, cfg.components_js, {
+      :size => cfg.max_renderers, :timeout => cfg.timeout, :json_engine => ::Oj
+    })
+
+    result = React::Renderer.render "TodoList", :todos => %w{todo1 todo2 todo3}
+    assert_match(/todo1.*todo2.*todo3/, result)
+    assert_match(/data-react-checksum/, result)
```
rmosolgo: Is expecting the same rendered output as the test below? If so, how can I tell that `Oj` (not `JSON`) was used to create this output?

One way to be sure would be to make a dummy JSON library, for example 

```
class StupidJSON 
  def self.encode(*args)
    %{ { "some_key" : "some_value" } }
  end
end
```

Then use that dummy encoder and make sure you get the expected output.

rmosolgo:
Cool feature! Would be nice to have a test that would catch any regression. Don't mind the failures on Travis, we have issues with our `coffee-script-source` version :(

test/react_renderer_test.rb
mickey: :+1: I update the PR with your recommendation.

mickey:
Hey @rmosolgo! Thanks for your review.
I changed the test to one using a custom dumb encoder.

vipulnsward:
This can be configured on rails config itself. Having two separate engines would cause confusion and unusual problems. 
We are doing custom dump/encode, which is wrong way for handling this, since rails encoder uses `as_json` with new design, and there may be incompatibilities doing direct calls to dump/encode wrt old/new engine. 
Also, doesn't `Oj`/`oj_mimic_json` already take care of this, if bundled in rails Gemfile?

rmosolgo:
good call @vipulnsward , just using the object's `#to_json` is a good OOP and Rails-y behavior. 

@mickey does that work in your case?

mickey:
@vipulnsward @rmosolgo Thanks for your feedback. It sure works for me and makes total sense.

Issue #108, opened 2014-11-05
Possible bug on Turbolinks


maxguzenski:
Hi,
Im trying to implement a support to pjax, when I see this on code:

```javascript
typeof Turbolinks !== 'undefined' ? handleTurbolinksEvents() : handleNativeEvents();
```

But Turbolinks sometime execute a fullpage reload:
* Some links (\<a/\>) can force a full page reload
* If turbolinks detect a change on assets, it will force a full page reload
* If turbolinks detect a timeout, it will force a full page reload

So, handleNativeEvents should not be required anyway? (with or without turbolinks) ??

rmosolgo:
@maxguzenski  took a pass at this in #183, what do you think of that?

rmosolgo:
We don't unmount on beforeUnload anymore, so I think this is no longer relevant!

PR #183, opened 2015-02-23
fix(ujs) always handle native events, _also_ handle turbolinks if appropriate

rmosolgo:
As mentioned in #108 , sometimes a turbolinks app still has native events. So, 

- always handle native events 
- also support turbolinks events 

Since the UJS functions are public now, we could put turbolinks UJS in its own file, right? Using turbolinks would be:

```js
//= require react_ujs
//= require react_turbolinks_ujs // or something
```

That's a backwards incompatibility though, maybe we should wait a bit

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -79,6 +79,9 @@
     }
     handleEvent('page:change', window.ReactRailsUJS.mountComponents);
     handleEvent(unmountEvent, window.ReactRailsUJS.unmountComponents);
+    // you can get a native page-change a couple ways, eg after deploying new assets, clicking data-no-turbolink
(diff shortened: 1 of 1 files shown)


xionon:
> That's a backwards incompatibility though, maybe we should wait a bit

I don't know if it would divide well along this line, but maybe you could have `react_native_ujs` and `react_turbolinks_ujs`, and have `react_ujs` just be a manifest that includes both?

rmosolgo:
oh @xionon good call, that would be a nice way of splitting it. I may try it in a later PR.

rmosolgo:
just realized this could result in an unexpected double-mount on first page load, thinking...

rmosolgo:
Ok, took another try at this. It should just listen for native unload even when doing turbolinks events.

rmosolgo:
we don't do anything on `unload` anymore

Issue #175, opened 2015-02-17
initializer react_rails.setup_vendor fails if gem source files were read-only


jonswar:
These lines in rails/railtie.rb:

        FileUtils.cp(::React::Source.bundled_path_for(filename),
                     tmp_path.join('react.js'))
        FileUtils.cp(::React::Source.bundled_path_for('JSXTransformer.js'),
                     tmp_path.join('JSXTransformer.js'))

will, the first time, create read-only files if the gem sources themselves are read-only (as they are on some deployed systems). On subsequent runs, the copy will fail, causing the initializer to error out. This code should either chmod the files or not copy them if they are already there.

danielrhodes:
@jonswar Having the same issue. I'll make a PR and see if the chmod/check if exists works.

rearadmiral:
Just noticed we still have some hacks in our Rails env file to workaround this. @danielrhodes did you ever have a chance to work on a PR for this? If not, I could take a look.

danielrhodes:
@rearadmiral Yep done. Didn't do anything fancy. Has been working for a couple months in production.

rearadmiral:
Awesome. Thanks

rmosolgo:
Good news: #254 please give it a try!

Issue #155, opened 2015-01-29
React components not reloading when testing


mockdeep:
I'm not sure if this is an issue with `react-rails` or `spring`, or something else, but I'm having a problem that when I use `spring rspec`, my react components aren't reloaded when I change them in the specs, so I have to run `spring stop` or just go without spring entirely.

rmosolgo:
For my part, I suspect our initializers are a bit off. The problem is, I don't ever have this problem (or other possibly-initializer-related problems). Would you mind testing my hypothesis?

You could point your gemfile at the branch in #187 and see if the problem still happens:

```ruby
gem 'react-rails', github: 'rmosolgo/react-rails', ref: '0baaff7fb67010295630af1bd21bf3749cb64ed1'
```

I think that would do the trick! I'd love to hear if you can try it and if you see any difference.

mockdeep:
Hey @rmosolgo. That doesn't seem to fix the issue for me. However, I have discovered something else. Whether or not it reloads properly seems to depend on the location of the files. If I change a file in `/app/assets/javascripts/` it reloads as expected. However if it's nested one level deeper in `/app/assets/javascripts/components/` it does not reload. In both cases I'm editing a `.jsx` file.

rmosolgo:
do you mind sharing your `Gemfile.lock`? At least for rails version and react-rails version if you don't want to post the whole thing

mockdeep:
Sure, no problem:

```
GIT
  remote: git://github.com/reactjs/react-rails.git
  revision: ac4320eb183b948ec328a695cd54d27c769073af
  specs:
    react-rails (1.0.0.pre)
      coffee-script-source (~> 1.9)
      connection_pool
      execjs
      rails (>= 3.1)
      react-source (~> 0.12)

GEM
  remote: https://rubygems.org/
  remote: https://rails-assets.org/
  specs:
    actionmailer (4.2.0)
      actionpack (= 4.2.0)
      actionview (= 4.2.0)
      activejob (= 4.2.0)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 1.0, >= 1.0.5)
    actionpack (4.2.0)
      actionview (= 4.2.0)
      activesupport (= 4.2.0)
      rack (~> 1.6.0)
      rack-test (~> 0.6.2)
      rails-dom-testing (~> 1.0, >= 1.0.5)
      rails-html-sanitizer (~> 1.0, >= 1.0.1)
    actionview (4.2.0)
[...]

Issue #148, opened 2015-01-17
TypeError: React.addons is undefined


tirdadc:
I occasionally and inconsistently get this error, which ends up breaking my component:

<code>TypeError: React.addons is undefined</code>

my use of React.addons is limited to this in this component to control classes:

<code>var cx = React.addons.classSet;</code>

and I do have <code>config.react.addons = true</code> in my development.rb. Restarting the Rails application seems to fix the issue, but obviously that's not going to be a solution.

mattmarcello:
I will add that I've encountered this same problem in a production environment.  There was no issue in development, however.

blainekasten:
Try putting the `config.react.addons = true` in the `config/application.rb` file and see if that still works.

rmosolgo:
For my part, I suspect our initializers are a bit off. The problem is, I don't ever have this problem (or other possibly-initializer-related problems). Would you mind testing my hypothesis?

You could point your gemfile at the branch in #187 and see if the problem still happens:

```
gem 'react-rails', github: 'rmosolgo/react-rails', ref: '0baaff7fb67010295630af1bd21bf3749cb64ed1'
```

I think that would do the trick! I'd love to hear if you can try it and if you see any difference.

KendallPark:
I am having this issue as well. I have `config.react.addons = true` set in my `application.rb` file. `addons` is not defined on `React`.

KendallPark:
Nm. Forgot I was using the 0.13 RC. Grabbed the right file with the addons and it's working as expected.

tirdadc:
@rmosolgo will try that and report back!

rmosolgo:
please reopen if you encounter this again with the latest version!

tirdadc:
Just chiming in to say that this seems to have gone away with the latest version. Thanks again for the stellar work, guys!

hopsoft:
__Note:__ `config.react.addons = true` must be placed in `application.rb`. Placing it in an initializer does not work.

paulkaplan:
The readme makes it sound like it belongs in `environment/[production|development].rb`.

trivektor:
I'm encountering this in 1.4.2.

PR #187, opened 2015-02-26
fix(Railtie) make sure initializers always run

rmosolgo:
It seems like we have a problem with initializers:

- not reloading when testing #155
- "Addons is undefined" #148 
- react source files aren't copied #175

Messing with the initializers has been recommended before: 

- changing the `group:` #40 
- initialize after sprockets #36 (However that initializer [no longer exists](https://github.com/rails/sprockets-rails/issues/227))

But I couldn't find much about how initializers run, just an [offhand mention in the Railtie docs](http://edgeapi.rubyonrails.org/classes/Rails/Railtie.html) (do you know of anything better?). 

I did find a couple of examples of `group: :all` in the wild:

- [sass-rails does it](https://github.com/rails/sass-rails/blob/master/lib/sass/rails/railtie.rb)
- [ember-rails does it](https://github.com/emberjs/ember-rails/blob/master/lib/ember_rails.rb)

I think it's worth a try, you?

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -16,21 +16,19 @@ module React
       config.react.component_filenames = ['components.js']
 
       # Watch .jsx files for changes in dev, so we can reload the JS VMs with the new JS code.
-      initializer "react_rails.add_watchable_files" do |app|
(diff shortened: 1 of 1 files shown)


vipulnsward:
LGTM

Issue #159, opened 2015-02-01
react_component not rendering on browser back navigation


rringham:
I've tried a few different things, but I'm unable to get my React components to render when navigating using the browser back button (sometimes even the forward button).

It doesn't seem to matter whether I prerender them server-side or if I render them client-side. E.g., this:

<%= react_component('TopicLinker', {topicName: @topic.name, topicId: @topic.slug}, {prerender: true}) %>

The entire mount node just doesn't show up in the DOM when navigating using the browser back button - multiple browsers, too - Safari (latest), Chrome (latest), etc - so I don't think this is an issue with React.js itself - feels more like it has to do with react-rails.

Thoughts? I know this is vague, but I'm really not sure what I'm doing wrong (only thing I can seem to think is that Turbolinks might be breaking this somehow). Everything else with react-rails works perfect, except with browser back navigation.

rringham:
Did some more testing, and it looks like it comes down entirely to Turbolinks. If I disable Turbolinks, the problem goes away.

I'd prefer to not disable Turbolinks... but I'm worried that this is going to be a limitation of react-rails.

rringham:
Looks like this resolves this for Safari - it's combination of disabling Turbolinks, and the Safari window.onunload trick:

```
<body data-no-turbolink data-no-transition-cache>
...
<script type="text/javascript">
    window.onunload = function(){};
</script>
</body>
```

rmosolgo:
Yes, I think some versions of react-rails + some versions of Turbolinks have this bug: 

- when a page is restored from the Turbolinks cache, components aren't mounted

I think the solutions are: 

- Disable the turbolinks cache with `Turbolinks.pagesCached(0)`
- use react-rails `master` + the latest version of Turbolinks

See related #89

luopio:
I had similar issues and my solution was to remove the `data-react-class` attribute in the component rendered by `react_component`. Something along these lines inside the component:
```javascript
componentDidMount: function() {
  $(this.getDOMNode()).parent().removeAttr('data-react-class');
},
```
The root of the issue seemed to be Turbolinks saving page state when leaving the page and after component unmounting in rails_ujs. I'm not sure what kind of long-term issues this might cause but seems to work so far.

johnthethird:
[...]

PR #190, opened 2015-03-01
fix(ujs) don't remove nodes after unmounting components

rmosolgo:
Don't remove wrappers after unmounting nodes because it breaks Turbolinks cache (#159). 

I can't find the PR where this was added, I would like to go back and see why, but now I can't find it :( 

Anyways, I was surprised this broke because I remember we have a test for navigating back and forth and making sure things mount ok. I added a test and found that in fact, Turbolinks isn't supported in the test environment: 

```  ruby
test 'turbolinks works in test'  do
  assert(page.evaluate_script('Turbolinks.supported')) # fails
end
```

So, the test doesn't guarantee turbolinks works :(

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -47,8 +47,6 @@
         var node = nodes[i];
 
         React.unmountComponentAtNode(node);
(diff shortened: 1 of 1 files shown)


luopio:
Works here. I can get rid of my `data-react-class` removing code and back navigation still works.

PR #191, opened 2015-03-01
chore(docs) reorganize readme

rmosolgo:
It seemed to me that the readme had grown paragraph-by-paragraph for quite some time and was due for some reorganization :)

I hope that more structure and shorter overall length will help users find what they are looking for. My approach was:

- Group information by feature
  - Put configurations near the feature they configure
  - For nearly all users, UJS & view helper are closely linked, so put them close to each other
  - Add table-of-contents-like links at the top of the readme
- Increase content related to FAQs, decrease less-recently-requested info
  - prerender issues, eg no `document`, global scope required, include your dependencies
  - lots of mentions of restarting your dev server 
  - shrink docs related to drop-in copies of JS libs

You can see a rendered version of it here: https://github.com/rmosolgo/react-rails/blob/readme-shuffle/README.md

What do you think? Is this an improvement? Have I shorted anything or given too much space to anything?

I think this is more organized but if you disagree I won't push through on it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,9 @@ MyApp::Application.configure do
   # array of filenames that will be requested from the asset pipeline
   # and concatenated:
   config.react.component_filenames = ['components.js']
+  # server-side console.log, console.warn, and console.error messages will be replayed on the client
+  # (you can set this to `true` in config/enviroments/development.rb to replay in development only)
+  config.react.replay_console = false
 end
 ```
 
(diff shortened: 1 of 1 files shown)


rstudner:
This needs another edit.

It mentions, up front & correctly to do the rails g install script which puts react/react_ujs & components.js in the application.js file.

In then, a page later, has the very old doc that says to just put solely and only react.js in the application.js (manually).

rmosolgo:
Thanks for your review @rstudner ! I've removed those suggestions of manually requiring things and made the details of the `install` task a bit more obvious, what do you think?

rmosolgo:
rebased and updated for `replay_console` config

PR #194, opened 2015-03-03
fix(CI) specify higher nodejs version

rmosolgo:
I'm a simple man, I just want a green badge :weary: 

I read on the coffee-script issue that one guy had better luck with node 0.12. I'm not exactly sure how to declare it with multiple ruby versions, but let's see...


rmosolgo:
no :game_die: :(

Issue #133, opened 2014-12-18
Ship 1.0


zpao:
Let's use this to discuss and track what needs to be done so that we can ship 1.0.

- [x] Figure out versioning scheme
- [x] Better docs / examples
- [ ] *your idea here*

uberllama:
Flux. Or Reflux. Please and thank you. ;)

rmosolgo:
@zpao  I think we're on the home stretch :D

- For my part, I'm happy with the docs after #191
- Last UJS issue is in #221, I'll merge it soon if nobody objects
- Starting the conversation about versioning in #228 

Any other blockers that you know of?

zpao:
Nah, take us home!

If anything comes up, just bump versions appropriately (I'm much more on board with that approach than I used to be).

rmosolgo:
@zpao of course there's still a few outstanding issues, but it's good to be `1.0.0` by me, what do you think?

zpao:
sounds like you've made a ton of progress and it's all your call! :shipit:

chantastic:
Great work @rmosolgo!

rmosolgo:
:tada: https://rubygems.org/gems/react-rails/versions/1.0.0

chantastic:
:clap: :clap: :clap:

zackify:
:+1:

jordanstephens:
🚀

jakegavin:
:confetti_ball: Nice work!

PR #195, opened 2015-03-04
Little change on note

andreynering:


rmosolgo:
image

Interesting, care to share your motivation?

andreynering:
I didn't note this warning myself the first time I read the docs, and ended using the master branch in production.

A warning like this should be big, so users note it.

Also, isn't the master branch done to be published in rubygems.org? It seems stable to me.

rmosolgo:
oh, i see, that is bigger, thanks! 

regarding 1.0, feel free to chime in on #133!

Issue #122, opened 2014-11-25
Console.log From Server Side Rendering


jadmz:
Hi,

Is there a way to print to the terminal while rendering from the server with jruby and ruby rhino? I would like to use it for debugging.

Thanks,
Joe

bikegriffith:
+1, would be tremendously helpful

rearadmiral:
Has anyone been able to get js stack traces? I only ever see the error's message and when it's something generic it makes it really tough to track down a problem.

bikegriffith:
@rearadmiral I left the stack trace out intentionally.  With a little bit more work, I could update it to include a representation of the stack as an additional part of the message string, but I was really trying to just get *something* better than what we have now.  Seems like a future enhancement rather than part of this request unless others want it now.

rmosolgo:
The error message includes the ExecJS stack trace since #215, horray!

PR #196, opened 2015-03-05
Replay console logging on client after server prerender

bikegriffith:
Addresses reactjs/react-rails#122.  Adds a new `replay_console: true` option, to be used in conjunction with `prerender: true` on the `react_component` view helper (can assist in debugging).  Updated README and added integration tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ end
 
 ### Server Rendering
 
-For performance and thread-safety reasons, a pool of JS VMs are spun up on application start, and the size of the pool and the timeout on requesting a VM from the pool are configurable. You can also say where you want to grab the `react.js` code from, and if you want to change the filenames for the components (this should be an array of filenames that will be requested from the asset pipeline and concatenated together.)
+For performance and thread-safety reasons, a pool of JS VMs are spun up on application start, and the size of the pool and the timeout on requesting a VM from the pool are configurable.
 
 ```ruby
 # config/environments/application.rb
@@ -316,9 +316,14 @@ MyApp::Application.configure do
   config.react.timeout = 20 #seconds
   config.react.react_js = lambda {File.read(::Rails.application.assets.resolve('react.js'))}
   config.react.component_filenames = ['components.js']
+  config.react.replay_console = false
 end
-
 ```
+
+Other configuration options include:
+* `react_js`: where you want to grab the javascript library from
+* `component_filenames`: an array of filenames that will be requested from the asset pipeline and concatenated together
+* `replay_console`: additional debugging by replaying any captured console messages from server-rendering back on the client (note: they will lose their call stack, but it can help point you in right direction)
 
 ## CoffeeScript
 
--- a/lib/react/console.rb
+++ b/lib/react/console.rb
@@ -1,0 +1,30 @@
+module React
+  class Console
+    def self.polyfill_js
+      # Overwrite global `console` object with something that can capture messages
+      # to return to client later for debugging
+      <<-JS
+      var console = { history: [] };
+      ['error', 'log', 'info', 'warn'].forEach(function (fn) {
+        console[fn] = function () {
+          console.history.push({level: fn, arguments: Array.prototype.slice.call(arguments)});
+        };
+      });
+      JS
+    end
+
+    def self.replay_as_script_js
+      <<-JS
+      (function (history) {
+        if (history && history.length > 0) {
+          result += '\\n<scr'+'ipt>';
+          history.forEach(function (msg) {
+            result += '\\nconsole.' + msg.level + '.apply(console, ' + JSON.stringify(msg.arguments) + ');';
+          });
+          result += '\\n</scr'+'ipt>';
+        }
+      })(console.history);
+      JS
+    end
+  end
+end
--- a/lib/react/renderer.rb
+++ b/lib/react/renderer.rb
@@ -12,10 +12,11 @@ module React
 
     cattr_accessor :pool
 
-    def self.setup!(react_js, components_js, args={})
+    def self.setup!(react_js, components_js, replay_console, args={})
       args.assert_valid_keys(:size, :timeout)
       @@react_js = react_js
       @@components_js = components_js
+      @@replay_console = replay_console
       @@pool.shutdown{} if @@pool
       reset_combined_js!
       default_pool_options = {:size =>10, :timeout => 20}
@@ -43,9 +44,11 @@ module React
     def render(component, args={})
       react_props = React::Renderer.react_props(args)
       jscode = <<-JS
-        function() {
-          return React.renderToString(React.createElement(#{component}, #{react_props}));
-        }()
+        (function () {
+          var result = React.renderToString(React.createElement(#{component}, #{react_props}));
+          #{@@replay_console ? React::Console.replay_as_script_js : ''}
+          return result;
+        })()
       JS
       context.eval(jscode).html_safe
     rescue ExecJS::ProgramError => e
@@ -56,22 +59,15 @@ module React
     private
 
     def self.setup_combined_js
-      <<-CODE
+      <<-JS
         var global = global || this;
         var self = self || this;
         var window = window || this;
-
-        var console = global.console || {};
-        ['error', 'log', 'info', 'warn'].forEach(function (fn) {
-          if (!(fn in console)) {
-            console[fn] = function () {};
-          }
-        });
-
+        #{React::Console.polyfill_js}
         #{@@react_js.call};
         React = global.React;
         #{@@components_js.call};
-      CODE
+      JS
     end
 
     def self.reset_combined_js!
--- a/test/dummy/app/controllers/server_controller.rb
+++ b/test/dummy/app/controllers/server_controller.rb
@@ -2,4 +2,23 @@ class ServerController < ApplicationController
   def show
     @todos = %w{todo1 todo2 todo3}
   end
+
+  def console_example
+    hack_replay_console_config true
+    @todos = %w{todo1 todo2 todo3}
+  end
+
+  def console_example_suppressed
+    hack_replay_console_config false
+    @todos = %w{todo1 todo2 todo3}
+  end
+
+  private
+  def hack_replay_console_config(value)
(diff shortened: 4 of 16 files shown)


bikegriffith:
Hoping dc08ee4 also fixes the master build failure.

test/server_rendered_html_test.rb
```diff
@@ -16,22 +16,55 @@ def wait_to_ensure_asset_pipeline_detects_changes
...
       assert_match(/Updated/, response.body)
     ensure
       # if we have a test failure, we want to make sure that we revert the dummy file
+      FileUtils.mv file_without_updates, app_file
+      FileUtils.touch app_file
       wait_to_ensure_asset_pipeline_detects_changes
+    end
+  end
+
+  test 'react server rendering shows console output as html comment' do
```
xionon: I think, instead of doing the fragile "copy file dance," this test should just rely on a new route and jsx file in the test app. Even something super specific, like `/console_example`.

xionon:
Overall, I really like this PR.

test/server_rendered_html_test.rb
bikegriffith: Can do.

lib/react/renderer.rb
```diff
@@ -22,9 +23,9 @@ def self.setup!(react_js, components_js, args={})
       @@pool = ConnectionPool.new(default_pool_options.merge(args)) { self.new }
     end
 
-    def self.render(component, args={})
+    def self.render(component, args={}, replay_console=false)
```
rmosolgo: Could `replay_console` be a member of the `args` hash? That would be nice because it would keep the Ruby idiom of options-hash-last.

I think you could fetch it out with: 

```ruby
replay_console =  !!args.delete(:replay_console)
```

lib/react/console.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def self.replay_as_script_js
+      <<-JS
+      (function (history) {
+        if (history && history.length > 0) {
+          result += '\\n<scr'+'ipt>';
+          history.forEach(function (msg) {
+            result += '\\nconsole.' + msg.level + '.apply(console, ' + JSON.stringify(msg.arguments) + ');';
+          });
+          result += '\\n</scr'+'ipt>';
```
rmosolgo: Interesting about joining the script tag string ... what is that preparing for? I haven't seen it before!

lib/react/renderer.rb
bikegriffith: Thanks for feedback. I wanted to keep it out of `args` because that's meant for use by the React component directly. The other place that seems reasonable would be on the app.config.react object, then passing into `Renderer.setup!`

rmosolgo: A config would be good by me, then you could just turn it always-on in development and always-off in production!

As for mixing args, it's alright in my book by way of precedent in the [view helper](https://github.com/reactjs/react-rails/blob/master/lib/react/rails/view_helper.rb#L20-L21)

lib/react/renderer.rb
```diff
@@ -40,12 +41,14 @@ def context
...
-    def render(component, args={})
+    def render(component, args={}, replay_console=false)
       react_props = React::Renderer.react_props(args)
       jscode = <<-JS
-        function() {
-          return React.renderToString(React.createElement(#{component}, #{react_props}));
-        }()
+        (function () {
+          var result = React.renderToString(React.createElement(#{component}, #{react_props}));
+          #{replay_console ? React::Console.replay_as_script_js : ''}
```
vipulnsward: Won't this cause issues, if arguments in console history loose their context?

test/dummy/app/views/server/console_example.html.erb
```diff
@@ -0,0 +1 @@
+<%= react_component "TodoListWithConsoleLog", {:todos => @todos}, {:prerender => true, :replay_console => true} %>
```
vipulnsward: These can be changed to use ruby 1.9 hashes.

test/dummy/app/views/server/console_example_suppressed.html.erb
```diff
@@ -0,0 +1 @@
+<%= react_component "TodoListWithConsoleLog", {:todos => @todos}, {:prerender => true, :replay_console => false} %>
```
vipulnsward: Same with this and other places

test/dummy/config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Dummy::Application.routes.draw do
   resources :pages, :only => [:show]
-  resources :server, :only => [:show]
+
+  get 'server/console_example' => 'server#console_example'
+  get 'server/console_example_suppressed' => 'server#console_example_suppressed'
+  get 'server/:id' => 'server#show'
```
vipulnsward: You can change this to:
```ruby
resources :server, only: [:show] do 
  collection do 
    get :console_example
    get :console_example_suppressed
  end
end
```

README.md
```diff
@@ -221,6 +221,8 @@ react_component('HelloMessage', {name: 'John'}, {prerender: true})
 ```
 This will return the fully rendered component markup, and as long as you have included the `react_ujs` script in your page, then the component will also be instantiated and mounted on the client.
 
+For additional debugging, you can optionally include `replay_console: true` in the options hash to replay any captured console messages from server-rendering back on the client.
+
```
vipulnsward: You can expand on saying, we are just adding as comments instead of actually running the console calls.

lib/react/renderer.rb
xionon: :+1: making it a config option

test/dummy/config/routes.rb
rmosolgo: :+1: for idiomatic routes :)

lib/react/renderer.rb
rmosolgo: I think this is addressed by the `JSON.stringify` https://github.com/reactjs/react-rails/pull/196#discussion-diff-25880836R22

rmosolgo:
Great feature! This will be a big win for day-to-day happiness!

lib/react/console.rb
bikegriffith: A straight-up "<script>" (or "</script>") seen anywhere can break the flow of execution in the JS interpreter so need to be careful.

lib/react/renderer.rb
bikegriffith: Perhaps, but it's better than what we have now, which is nothing.

bikegriffith:
Thanks @xionon, @rmosolgo, and @vipulnsward for all the feedback and getting this into better shape.  Hoping this can make it into master. I've been using this branch locally and it's been helpful for my workflow.

rmosolgo:
Changes look great! :+1:

README.md
```diff
@@ -316,10 +316,15 @@ MyApp::Application.configure do
...
   config.react.component_filenames = ['components.js']
+  config.react.replay_console = false
 end
-
 ```
 
+Other optional configuration values include:
+* `react_js`: where you want to grab the javascript library from
+* `component_filenames`: an array of filenames that will be requested from the asset pipeline and concatenated together
+* `replay_console`: additional debugging by replaying any captured console messages from server-rendering back on the client (note: they will lose their call stack, but it can help point you in right direction)
```
vipulnsward: `NOTE`

README.md
```diff
@@ -316,10 +316,15 @@ MyApp::Application.configure do
   config.react.timeout = 20 #seconds
   config.react.react_js = lambda {File.read(::Rails.application.assets.resolve('react.js'))}
   config.react.component_filenames = ['components.js']
+  config.react.replay_console = false
 end
-
 ```
 
+Other optional configuration values include:
```
vipulnsward: `configuration options`

lib/react/renderer.rb
```diff
@@ -1,4 +1,5 @@
 require 'connection_pool'
+require_relative './console'
```
vipulnsward: This can go at top-level require in `lib/react-rails.rb`

test/dummy/app/controllers/server_controller.rb
```diff
@@ -2,4 +2,21 @@ class ServerController < ApplicationController
   def show
     @todos = %w{todo1 todo2 todo3}
   end
+  def console_example
```
vipulnsward: newlines after methods

test/dummy/app/views/server/show.html.erb
```diff
@@ -1 +1 @@
-<%= react_component "TodoList", {:todos => @todos}, :prerender => true %>
+<%= react_component "TodoList", {:todos => @todos}, {:prerender => true} %>
```
vipulnsward: missed this one.

vipulnsward:
@bikegriffith added more comments. Looking forward to this. :smile:

rmosolgo:
looks good to me, last call @vipulnsward @xionon ?

vipulnsward:
LGTM . @bikegriffith can you squash the commits? Its good to track this as a single change to refer in future.

bikegriffith:
@vipulnsward I attempted to squash the commits and also pulled the latest from upstream.  Let me know if I did anything weird and I'll try to correct.  Thanks!

vipulnsward:
Ah, I am afraid an extra commit got in. But that should be ok, since https://github.com/bikegriffith/react-rails/commit/585963f81181891bfb2f64c44e1ad103a67a097e is a clean commit.
So :+1:

rmosolgo:
Just got a chance to try this out locally. (It took some finagling, my app isn't prerender-ready!). 

I noticed this outputs a warning, too:

image

But this seems OK with me, since it's meant for development only (right?).

bikegriffith:
@rmosolgo I only intended it to be used for development, so I'm less concerned, but it is strange because I wasn't seeing it. Are you on ReactJS 0.13?

rmosolgo:
Yep, that was 0.13. I checked out your branch so I'm not sure how that snuck in there! Maybe because react-rails/master has 0.13 now

rmosolgo:
:confetti_ball: thanks!

PR #200, opened 2015-03-10
v0.13.0.0

zpao:
:exclamation: don't merge me :exclamation: 

I'll merge this directly so we have the right branches set up. This will branch from the 0.12 branch.


zpao:
@rmosolgo Wanted to give you a heads up before I pushed this. This does not impact master. I'll let you handle that as you see fit.

zpao:
Ah you commented in the issue. I pushed this as 4da2126f948a934b2c56da9d9b1ce653491e41a6 into a new 0.13 branch based on 0.12 and tagged appropriately.

PR #208, opened 2015-03-13
Inline props operations improvement

panasyuk:
# Problem
1. Size of inline props placed to the root component tag as attributes may be too huge to:
  1.1 Process in browser devtools. So that browser stucks trying to render markup for preview.
  1.2 Deliver it to user due to bandwidth (for highload apps).
2. Inline props placed to the attributes of tag are looking freaky because of its size and HTML safety.

So props we want to place:
```JSON
{"foo": "bar"}
```
are looking like this:
```HTML
{&quot;foo&quot;:&quot;bar&quot;}
```

# Solution
Moving inline props to the ```<script type="text/json">{"foo": "bar"}</script>``` will:
1. Reduce the size of data passed to the page in order to render component at the initial state. It's also useful because it reduces bandwidth utilization (even if we are using GZIP).
2. Allow browser inspector to operate on DOM faster.
3. Make markup looking more cleaner.

Placing inline props at the end of body speeds up DOM rendering and allows Turbolinks to load this data along with other markup on each request.

# How it works
The process of placing the components to the markup in Rails is still the same except of developer should add ```<%= render_react_props %>``` at the end of ```<body>``` tag.

# P.S.
We selected coffee-script-source at 1.8 because of [this issue](https://github.com/jashkenas/coffeescript/issues/3829).

### DIFF
--- a/lib/react/rails/view_helper.rb
+++ b/lib/react/rails/view_helper.rb
@@ -9,13 +9,15 @@ module React
       def react_component(name, args = {}, options = {}, &block)
         options = {:tag => options} if options.is_a?(Symbol)
         block = Proc.new{concat React::Renderer.render(name, args)} if options[:prerender]
+        separate_props = options.delete :separate_props
+        move_separate_props_out = options.delete :move_separate_props_out
 
         html_options = options.reverse_merge(:data => {})
         html_options[:data].tap do |data|
           data[:react_class] = name
           next if args.empty?
-          if react_separate_props?
-            data[:react_props_id] = add_react_props args, options[:inline_props]
+          if separate_props
+            data[:react_props_id] = add_react_props args, move_separate_props_out
           else
             data[:react_props] = React::Renderer.react_props args
           end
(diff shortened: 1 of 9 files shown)


react-rails.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'jbuilder'
 
   s.add_dependency 'execjs'
-  s.add_dependency 'coffee-script-source', '~>1.9'
+  s.add_dependency 'coffee-script-source', '~>1.8'
```
vipulnsward: Its already been solved in master. Can you sync up.

panasyuk: Done

lib/react/rails/view_helper.rb
```diff
@@ -23,6 +23,26 @@ def react_component(name, args = {}, options = {}, &block)
         content_tag(html_tag, '', html_options, &block)
       end
 
+      # Add properties for component and return element id.
+      #
+      def add_react_props(props={})
+        return if props.empty?
+        props_id = SecureRandom.base64
+        content_for :react_props do
```
vipulnsward: This will make the props non-cachable.

vipulnsward:
I think we should make this an opt-in behavior. Add a react config and based on it, allow inline vs script props. 
Also now we need to call `react_component` as well as `render_react_props`. 
You need to fix test to handle it.

lib/react/rails/view_helper.rb
panasyuk: Which one element of application stack will lose ability to cache this?

vipulnsward: view caches

panasyuk:
@vipulnsward Please checkout last CI run. I implemented this with tests but for some versions tests are failing due to issue i wrote about.

Issue #209, opened 2015-03-13
Prerender fails with react and jQuery loaded from npm


janroesner:
In order to have CommonJS `require()` and JEST work, I followed the setup described [here](https://reactjsnews.com/setting-up-rails-for-react-and-jest/). So in `app/assets/javascripts` I created an `application.js` with:

```javascript
//= require components
//= require npm_jquery
```

Besides a `components.js` with:

```javascript
//= require_self
//= require react_ujs

React = require('react');
DemoComponent = require('./components/DemoComponent');
```

and a `npm_jquery.js` with:

```javascript
//= require_self

jQuery = require('jquery');
jQueryUJS = require('jquery-ujs')
```

When I render my component in my view with prerendering enabled like so:

```ruby
<%= react_component 'DemoComponent', { name: 'Jon Doe' }, { prerender: true } %>
```

I receive the following error:

**React::Renderer::PrerenderError in Pages#index**
Showing /Users/jondoe/react/rails/test-app/app/views/pages/index.html.erb where line #2 raised:

Encountered error "Error: Cannot find module 'function (document, window) {
  // jQuery is optional. Use it to support legacy browsers.
  var $ = (typeof window.jQuery !== 'undefined') && window.jQuery;

  // create the  namespace
  window.ReactRailsUJS = {
    CLASS_NAME_ATTR: 'data-react-class', ...

Rendering w/o `prerender` set works totally fine and yields expected results. Is this a problem with my setup, or did I encounter a bug?

janroesner:
Update: I found a workaround to make prerendering, JEST based testing and CommonJS `require` work, but I do not know, whether this causes problems, that I do not see at the moment. My current solution:

`app/assets/javascripts/application.js`

```javascript
//= require jquery
//= require jquery_ujs
//= require react_ujs
//= require components
```

`app/assets/javascripts/components.js`

```javascript
React = require('react');
// Explicitely ONLY require components that are used inside Rails helpers
// NOT SubComponents
DemoComponent = require('./components/DemoComponent');
```

In the `app/assets/javascripts/components` folder only components are placed, that are inserted into the page via the `react_component` helper. In here, CommonJS `require` statements work properly:

`app/assets/javascripts/components/DemoComponent.js.jsx`

```javascript
var React = require('react');
var uniq = require('uniq');
var SubComponent = require('../subcomponents/SubComponent');

var DemoComponent = React.createClass({

  render: function() {
[...]

PR #210, opened 2015-03-14
adding semi-colon in front of IIFE

orlando:
# Description

This PR address an edge case when there's bad written JavaScript before requiring `react_ujs`. Given how automatic semi-colon insertion works in JavaScript, an IIFE can be evaluated as a function call instead of a evaluation context.

This happened to me using `browserify-rails` and requiring `react_ujs` just after doing `require('React')`, but it can happen in multiple scenarios.

## Tasks

* [x] Add semi-colon before IIFE in **react_ujs.js.erb** file

## Screenshots

screen shot 2015-03-14 at 10 59 05 am

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -1,7 +1,7 @@
 /*globals React, Turbolinks*/
 
 // Unobtrusive scripting adapter for React
-(function(document, window) {
+;(function(document, window) {
   // jQuery is optional. Use it to support legacy browsers.
   var $ = (typeof window.jQuery !== 'undefined') && window.jQuery;
 


rmosolgo:
wow, I've seen those leading semicolons before but I never knew what they were for! 

(fancy that, I've had this kind of bug before (not react-rails, somewhere else) but I didn't know this was one way to prevent it)

Thanks!

vipulnsward:
@orlando I am curious, this order of require cause any other issues?
Usually you would require ujs first.

rmosolgo:
Also, this issue describes a similar error message: https://github.com/reactjs/react-rails/issues/209

orlando:
@vipulnsward It's ok, two or more semi-colons in the same statement is not an error.

For example
```(js)
var a = 2;;;;;
;;;;;(function () {
  console.log(a);
}());
```

orlando:
@rmosolgo Yeah, this PR should fix #209

vipulnsward:
@orlando yeah, I get the use case, wast curious about the order in which you required.
Requiring `react_ujs` before other requires wouldn't cause this issue.

PR #215, opened 2015-03-19
Include a stacktrace from execjs

andygeers:
The latest version of execjs now includes backtraces in Javascript errors - this makes debugging problems much easier!

### DIFF
--- a/lib/react/renderer.rb
+++ b/lib/react/renderer.rb
@@ -5,7 +5,8 @@ module React
 
     class PrerenderError < RuntimeError
       def initialize(component_name, props, js_message)
-        message = "Encountered error \"#{js_message}\" when prerendering #{component_name} with #{props}"
+        message = ["Encountered error \"#{js_message}\" when prerendering #{component_name} with #{props}",
+                    js_message.backtrace.join("\n")].join("\n")
         super(message)
       end
     end
--- a/test/react_renderer_test.rb
+++ b/test/react_renderer_test.rb
@@ -36,18 +36,11 @@ class ReactRendererTest < ActiveSupport::TestCase
   end
 
   test 'prerender errors include the error backtrace' do
-    # Stub the backtrace returned by ExecJS
-    ExecJS::ProgramError.class_eval {
-      def backtrace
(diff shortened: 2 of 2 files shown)


vipulnsward:
@andygeers tests are failing, could you take a look?
Also we support rubyracer, etc. Would this be affected by this?

rmosolgo:
Looks like the same failing test that's been bugging us all along:

```
  2) Error:
ViewHelperTest#test_react_server_rendering_does_not_include_internal_properties:
ActionView::Template::Error: TypeError: Cannot set property 'root' of null
  (in /home/travis/.rvm/gems/ruby-2.0.0-p598/gems/turbolinks-2.5.3/lib/assets/javascripts/turbolinks.js.coffee)
```

That lingering Turbolinks bug, tried to address here, https://github.com/reactjs/react-rails/pull/168 but locking to `1.8.0` gave people other problems.  

@andygeers could you add a test that checks for the presence of a backtrace? That would give some insight to how it's treated on the different JS runtimes (and make sure this feature keeps working!)

rmosolgo:
Sorry, should have said first: thanks, this will be a great feature!! :D

andygeers:
@rmosolgo I'm not sure if that's the kind of test you were looking for?
@vipulnsward From what I can tell, the change in execjs applies to all the different runtime engines they support - we're using rubyracer here, and there are definitely commits related to fixing it in rhino too

rmosolgo:
Yeah, like that, but preferably without stubbing `ExecJS::ProgramError#backtrace`. While the stub is present, we wouldn't know if the ExecJS API changed. 

I imagine it's hard (and pointless) to test the _actual_ content of the backtrace, but I think it would be an improvement even just to test for a bunch of newlines, for example, 

```ruby
assert(err.message.count("\n") > 2, "The error has a multi-line backtrace")
```

That puts a nice regression test on getting what we expect from ExecJS!

andygeers:
Ok, done

rmosolgo:
:+1: :heart_eyes_cat: it'll be great to actually get some context from the prerender environment!

rmosolgo:
thanks again!

PR #216, opened 2015-03-19
update to React 0.13.1

rmosolgo:
We already have a `0.13` branch that uses 13, but it's time to update master too I think


vipulnsward:
:100:

PR #221, opened 2015-03-24
Use a selector to mount and unmount components at a given node

jakegavin:
I'm working on an app where we need to be able to manually unmount specific components when a modal is closed. This PR adds a `searchSelector` argument to `ReactRailsUJS.mountComponents` and `ReactRailsUJS.unmountComponents` which lets you target specific elements. 

This is similar to https://github.com/reactjs/react-rails/pull/91. This functionality has been listed as a goal for 1.0 (https://github.com/reactjs/react-rails/issues/133) but the author of #91 hasn't responded in two months so I opened this PR.

### DIFF
--- a/test/view_helper_test.rb
+++ b/test/view_helper_test.rb
@@ -48,21 +48,21 @@ class ViewHelperTest < ActionDispatch::IntegrationTest
     assert html.include?('class="test"')
     assert html.include?('data-foo="1"')
   end
-  
+
   test 'ujs object present on the global React object and has our methods' do
     visit '/pages/1'
     assert page.has_content?('Hello Bob')
-  
+
     # the exposed ujs object is present
     ujs_present = page.evaluate_script('typeof ReactRailsUJS === "object";')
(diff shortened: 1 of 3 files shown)


danott:
:+1: I've found myself reaching for this exact behavior multiple times.

rmosolgo:
Nice feature! Looks good to me, let me just give it a try locally with Turbolinks and without

rmosolgo:
Looks good in a local test app w/Turbolinks & without, and looks good in PCO Check-ins

rmosolgo:
:tada: :confetti_ball:  thanks!

test/view_helper_test.rb
```diff
@@ -115,13 +115,24 @@ class ViewHelperTest < ActionDispatch::IntegrationTest
     assert page.has_content?('Hello Bob')
   end
 
+  test 'react_ujs can unount at node' do
```
vipulnsward: `unmount`

PR #228, opened 2015-03-30
feat(versions) add VERSIONS.md for gem/React versions

rmosolgo:
As expressed in #133, we need a way to better express how gem versions and react versions compare. 

I've taken a page from [jquery-rails](https://github.com/rails/jquery-rails)'s book and added a `VERSIONS.md` file with info about bundled versions. I also put information there about other ways of controlling the React version.

Right now, the update plan is pretty low-tech: I added comments to gem version and `react-source` version to remind anyone to update `VERSIONS.md`. In my opinion, this is better than what we have now and good enough for the time being. 

If you prefer, we could add a rake task for updating the react version. I think it's overkill (more maintenance overhead than the comments, more likely to break in the future), but jquery-rails does something similar and I'm open to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ end
 
 After restarting your Rails server, `//= require react`  will provide the build of React.js which was specified by the configurations.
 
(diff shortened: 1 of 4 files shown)


zpao:
:thumbsup:

rmosolgo:
Travis needs a "4/5 not bad" response :S

PR #230, opened 2015-03-30
refactor(Railtie) move react path calculation to separate class

rmosolgo:
`railtie.rb` has been growing bit-by-bit, I think we could clear it up a bit by moving this logic into another object.


zpao:
:thumbsup:

rmosolgo:
closed because copyless setup is better!

PR #237, opened 2015-04-07
Fix a documentation typo and several small proofing edits

christopherstyles:
There’s a typo in `VERSIONS.md` and a few small spots where the documentation can be a bit more consistent with colon usage and inline code spans (when referring to `react-rails` etc). Nothing major at all, just some tiny things I noticed :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 - Provide [various `react` builds](#reactjs-builds) to your asset bundle
 - Transform [`.jsx` in the asset pipeline](#jsx)
 - [Render components into views and mount them](#rendering--mounting) via view helper & `react_ujs`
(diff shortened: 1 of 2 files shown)


rmosolgo:
:+1: for proper parallel structure!

christopherstyles:
@rmosolgo anything I need to do to get this passing? 

It’s throwing `ActionView::Template::Error: TypeError: Cannot set property 'root' of null` – maybe the https://github.com/jashkenas/coffeescript/issues/3829 issue again?

rmosolgo:
Yes, that's the issue, sadly nothing to do that I know of!

rmosolgo:
thanks!

PR #239, opened 2015-04-11
Add jsx_transformer_js option

a2ikm:
It forces the path to JSXTransformer.js.

Currently `assets.resolve("JSXTransformer.js")` seems to search in `tmp/react-rails`, `vendor/assets/react` and `vendor/assets/react/development` at first.

So it's hard to place custom version in other directories like `vendor/assets/bower_components` managed with bower.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ MyApp::Application.configure do
   config.react.timeout = 20
   # where to get React.js source:
   config.react.react_js = lambda { File.read(::Rails.application.assets.resolve('react.js')) }
+  # where to get JSXTransformer.js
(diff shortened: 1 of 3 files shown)


rmosolgo:
@a2ikm thanks for taking a first pass at this, sorry some other changes have caused merge conflicts!

I took a stab at implementing this in #273, so you could pass it as a config option:
 
```ruby 
 config.react.jsx_transform_options = {
   asset_path: "bower_components/path/to/JSXTransformer.js" 
 }
```

What do you think of that?

rmosolgo:
I've merged #273, you can pass a custom `asset_path:` to `jsx_transform_options`!

a2ikm:
@rmosolgo Sorry for my late reply.  Your patch looks very nice, thank you!

Issue #240, opened 2015-04-14
Sprockets 3.0 support


kirs:
The Railtie in the gem fails under Sprockets 3.0 which was released recently:

```
/Users/kir/.rbenv/versions/2.0.0-p451/lib/ruby/gems/2.0.0/gems/react-rails-1.0.0/lib/react/rails/railtie.rb:49:in `block in <class:Railtie>': undefined method `prepend_path' for nil:NilClass (NoMethodError)
  from /Users/kir/.rbenv/versions/2.0.0-p451/lib/ruby/gems/2.0.0/gems/railties-4.2.1/lib/rails/initializable.rb:30:in `instance_exec'
  from /Users/kir/.rbenv/versions/2.0.0-p451/lib/ruby/gems/2.0.0/gems/railties-4.2.1/lib/rails/initializable.rb:30:in `run'
  from /Users/kir/.rbenv/versions/2.0.0-p451/lib/ruby/gems/2.0.0/gems/railties-4.2.1/lib/rails/initializable.rb:55:in `block in run_initializers'
```

I will work on the fix.

rmosolgo:
@kirs I'm trying to reproduce this locally -- what did you do to encounter this error? (I've gotta dummy app on Rails 4.2 + Sprockets 3.0.2 and it starts up & runs ok in development)

kirs:
Actually, now I found that I can reproduce it only with `sprockets-rails` on `master`.
If you don't consider it as a bug, feel free to close the issue.

rmosolgo:
ok, thanks, yeah I'll close this one and keep an ear out for sprockets 3 issues!

PR #241, opened 2015-04-14
Sprockets 3.0 support

kirs:
In Sprockets 3.0 `app.assets` is assigned in `after_initialize` callback (https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L142).
So we can use `app.assets` only from `after_initialize` in react-rails.

Solves #240

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -28,7 +28,7 @@ module React
         end
       end
 
-      initializer "react_rails.setup_vendor", group: :all do |app|
+      config.after_initialize do |app|
         # Mimic behavior of ember-rails...
         # We want to include different files in dev/prod. The unminified builds
         # contain console logging for invariants and logging to help catch
@@ -46,6 +46,7 @@ module React
                      tmp_path.join('react.js'))
         FileUtils.cp(::React::Source.bundled_path_for('JSXTransformer.js'),
                      tmp_path.join('JSXTransformer.js'))
+
         app.assets.prepend_path tmp_path
 
         # Allow overriding react files that are not based on environment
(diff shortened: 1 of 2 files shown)


kirs:
Apparently this patch doesn't work in test environment because `app.assets.paths` is frozen.

lib/react/rails/railtie.rb
```diff
@@ -46,17 +46,17 @@ class Railtie < ::Rails::Railtie
                      tmp_path.join('react.js'))
         FileUtils.cp(::React::Source.bundled_path_for('JSXTransformer.js'),
                      tmp_path.join('JSXTransformer.js'))
-        app.assets.prepend_path tmp_path
+        app.config.assets.paths.unshift(tmp_path)
```
kirs: @josh I couldn't use `config.assets.configure` here because it's `nil` and it's only set in `after_initialize`: https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L142

May I use operate directly on `app.config.assets.paths` here?

josh: Could you just wrap this entire "setup_vendor" thing in `config.assets.configure` instead of using an `initializer`?

Something like

``` ruby
config.assets.configure do |env|
  tmp_path = env.root.join('tmp/react-rails')
  # ...
  env.prepend_path tmp_path
end
```

kirs:
@rmosolgo any thoughts?

rmosolgo:
Humm, interesting fail. Seems like it's the same line that's causing this problem: https://github.com/reactjs/react-rails/issues/249 I guess we need to change the way we add things to the asset pipeline. 

@kirs one thing that would be really valuable would be to add a failing test. See how there's this Appraisal for a special sprockets version? https://github.com/reactjs/react-rails/blob/master/Appraisals#L5 We could add one for Rails 4 + Sprockets 3. That way we could be sure that, when we fix this, we _actually_ fix it :) 

I can take a deeper look at how react-rails uses the asset pipeline tomorrow or early next week. (Of course, if you get a chance, have at it!)

rmosolgo:
This recent build used sprockets 3 and turned out green, feel free to re-open if an incompatibility is discovered later on!

https://travis-ci.org/reactjs/react-rails/jobs/62607548#L341

PR #243, opened 2015-04-14
Rspec tests fail due to `tilt` dependency

gbanis:
I am using `react-rails` as dependency in a gem I created and running `rspec` in that gem throws the following error:

```
$ rspec
/Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react/jsx/template.rb:1:in `require': cannot load such file -- tilt (LoadError)
	from /Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react/jsx/template.rb:1:in `<top (required)>'
	from /Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react/jsx.rb:3:in `require'
	from /Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react/jsx.rb:3:in `<top (required)>'
	from /Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react-rails.rb:1:in `require'
	from /Users/gbanis/.rvm/gems/ruby-2.0.0-p598/gems/react-rails-0.13.0.0/lib/react-rails.rb:1:in `<top (required)>'
	...
```

This happens because `react-rails` requires `tilt` [here](https://github.com/reactjs/react-rails/blob/0.13/lib/react/jsx/template.rb).

I've noticed that adding `tilt` as a dependency in the `gemspec` solves this issue and all other tests run without a problem.

### DIFF
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'execjs'
   s.add_dependency 'rails', '>= 3.1'
   s.add_dependency 'react-source', '0.13.0'
-  spec.add_dependency 'tilt'
+  s.add_dependency 'tilt'
 
   s.files = Dir[
     'lib/**/*',
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'execjs'
   s.add_dependency 'rails', '>= 3.1'
   s.add_dependency 'react-source', '0.13.0'
+  spec.add_dependency 'tilt'
 
   s.files = Dir[
     'lib/**/*',


react-rails.gemspec
```diff
@@ -21,6 +21,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'execjs'
   s.add_dependency 'rails', '>= 3.1'
   s.add_dependency 'react-source', '0.13.0'
+  spec.add_dependency 'tilt'
```
zpao: `s`, not `spec`

zpao:
Can you actually do a PR against master as well? That makes sure it goes into the major releases.

gbanis:
@zpao Sure 
@facebook-github-bot All set

rmosolgo:
cool

zpao:
I can ship 0.13.2.0 with this and React 0.13.2. We could also do 0.13.1.1 if you want with just this.

gbanis-fiksu:
I'll make another PR to add this in 1.0.
Is there a reason why this doesn't get carried with the new versions?

rmosolgo:
No good reason, only poor maintaining on my part :grimacing:   Yeah, I just pushed the 1.0-stable branch for that purpose!

PR #244, opened 2015-04-15
docs(README): remove CoffeeScript doc bloc section

chantastic:
The doc bloc isn't required in the bundled version of React. I think this sentiment is no longer relevant.


rmosolgo:
thanks!

PR #250, opened 2015-04-23
Change permissions on copied file so subsequent deploys work.

### LINKED ISSUE #175: initializer react_rails.setup_vendor fails if gem source files were read-only

danielrhodes:
In some production environments, the permissions of the copied file were not being set in a way where subsequent deploys could overwrite the files if necessary. This sets the initial permissions of the two problematic files so that subsequent deploys do not run into this permissions issue.

Fixes #175

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -42,10 +42,18 @@ module React
                    (app.config.react.addons ? '-with-addons' : '') +
                    (app.config.react.variant == :production ? '.min.js' : '.js')
         FileUtils.mkdir_p(tmp_path)
+
+        # Copy over files and set permissions so subsequent deploys can overwrite files
         FileUtils.cp(::React::Source.bundled_path_for(filename),
-                     tmp_path.join('react.js'))
+                    tmp_path.join('react.js'))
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: I was going to say, can we add a test to make sure this behavior is preserved in the future, but I'm not sure where to add that test ... :S

rmosolgo:
@danielrhodes We just merged #254 which solves this at the root I think -- that PR removed the whole copy-on-boot workflow and replaced it by adding different paths to Sprockets based on your settings. I think that's a much cleaner flow.

Could you try on master and see if that works in your case?

rmosolgo:
Please let me know if this new solution doesn't work for your case!

Issue #143, opened 2015-01-06
Better support for Flux architecture


willcosgrove:
I've been working on a Rails app using React and Flux, and I've run into a few walls.  I thought I would check with the react-rails community before I put my pioneer hat on.

So let me run through really quickly how I *think* I would like things to work, and maybe someone can let me know if I'm way off.

## Serialization

*This is not really a concern of react-rails, but it's important to my overall plan*

I want all of my records for any given page serialized like so

```js
{
  posts: [
    {
      id: 1
      title: "There's always money in the banana stand!",
      body: "...",
      author_id: 2,
      comment_ids: [1]
    }
  ],
  comments: [
    {
      id: 1
      author_id: 1,
      body: "@gob you did mail that insurance letter, right?"
    }
  ],
  users: [
    {
      id: 1,
      name: "Michael Bluth"
    }, {
      id: 2,
      name: "George Bluth Sr."
    }
  ]
}
```

In other words: I want each model serialized in an array on the root object.  This will facilitate loading the models into their respective Flux stores.

## Stores

I want all of my data loaded into stores, and then I want my react components to be passed those stores through properties.  So rather than passing data directly into the component's properties, I want to tell react-rails which stores each component will need to access.  My first thought was to abandon the react-rails UJS and just initialize my components the old fashioned way, but I realized it would be a pretty trivial thing to add into the UJS.  This is what sparked the writing of this issue: I didn't know if this sounded like a good idea to anyone else, as I would be more than happy to submit a PR once it's all ready.

## Server Side Rendering

I'm not sure at the moment how to get this all working with server side rendering.  It's not a huge concern to me, but I realize that if this plan were to move forward at all with react-rails, there would need to be a way to get it to work with server side rendering.

My plan for getting the data into the stores was just going to be to pass them into [gon](https://github.com/gazay/gon) and then have a script grab everything out of gon, and throw them into the stores.  But gon wouldn't be available for server rendering, so I would need to work around that.  But I don't see that as being too difficult of an issue.

## Conclusion

If this sounds dumb, it's probably because it is.  I haven't actually tried to do any of the following yet, but it's my tentative action plan.  I would love to hear if anyone else has been successfully using the Flux architecture with react-rails, and what approach was taken.  I would also be happy to begin working on a better integration for Flux and react-rails if anyone is interested in this approach (or a better one :smile:).

rmosolgo:
[...]

PR #253, opened 2015-04-28
Decouple renderer

rmosolgo:
This is my attempt to decouple the server renderer as described in https://github.com/reactjs/react-rails/issues/143#issuecomment-76065541

- `React::ServerRendering` handles pooling & provides an interface to the rest of the gem
- It takes a renderer class who must implement `#initialize(options)` and `#render(component_name, props)`
- `React::ServerRendering::SprocketsRenderer` is an implementation which uses files from the asset pipeline to prerender components.

Whaddya think? Does this provide enough flexibility for issues like the #143? 

Also I added `guard` & `Guardfile`, I think it's a nice addition to the development flow, does anyone mind?


lib/react/rails/view_helper.rb
```diff
@@ -1,28 +1,28 @@
...
+
+        if options[:prerender]
+          block = Proc.new{ concat React::ServerRendering.render(name, props) }
+        end
 
         html_options = options.reverse_merge(:data => {})
         html_options[:data].tap do |data|
           data[:react_class] = name
-          data[:react_props] = React::Renderer.react_props(args) unless args.empty?
+          data[:react_props] = (props.is_a?(String) ? props : props.to_json)
```
rmosolgo: This should always be JSON because `ReactRailsUJS` will try to parse it as JSON!

willcosgrove:
This looks great!  I'll have to check it out tomorrow and see if I can use this API to build out something from my original example in #143.

rmosolgo:
Somehow I broke a lot of tests :disappointed:  gotta check that out.

johnthethird:
+1.  I would like to see a way to render with `renderToStaticMarkup` as well, would that be a totally different class like `SprocketsStaticRenderer` or an options flag `options[:static_markup] = true`?  I am leaning towards the option, as there are 2 ways to render on the server, and you should be able to switch between them.

rearadmiral:
As an abstraction layer, I think this looks good.

The direction I took my [implementation](https://github.com/goodreads/react-rails/blob/gr_prod/lib/react/renderer.rb#L76), basically to support seeding Flux stores from Rails data, reuses much of the impl of Renderer. It would be nice if there were some extension points to allow that.

For example:

```ruby
   js_code = <<-JS
      (function () {
        #{preparation_js}
        var result = React.renderToString(React.createElement(#{component_name}, #{props}));
        #{post_render_js}
        return result;
      })()
   JS

    ...

  def preparation_js
  end

  def post_render_js
    @replay_console ? CONSOLE_REPLAY : ""
  end
```

That way I could just subclass the Sprockets render and tweak a few things.

But maybe someone more clever than me can come up with a better way to handle the whole Flux thing anyway.

rmosolgo:
@johnthethird I made it so you can do `prerender: :static` and use `renderToStaticString` 

@rearadmiral I can jive with that, I'll do it tomorrow

rmosolgo:
I'm stumped on this failing test -- the asset pipeline serves the right file, but the route doesn't 

If i start up the dummy app, it works as expected:

image

rmosolgo:
Ugh, it's test order: if I remove the tests I added, everything runs ok.

rmosolgo:
Asking for help https://github.com/rails/sprockets/issues/51

rmosolgo:
It ain't pretty but it's working. 

My call to the asset pipeline [in SprocketsRenderer](https://github.com/reactjs/react-rails/pull/253/files#diff-1cd07a77af52dfe22d04099826f272a9R11) (in an earlier test) caused `react.js` to be cached. In an Sprockets version, it's not expired correctly, so I put a [workaround in the test](https://github.com/reactjs/react-rails/pull/253/files#diff-1aca847fb9fdaa5e0e82ed295e1c0fa6R21). `Asset#force?` doesn't exist in later Sprockets versions, so it's never called.

I'm bummed that we have a test-order issue :S  But, it's a very odd case, it only fails on old Sprockets, and the same actions work in the `dummy/` app if I boot it locally, so I'm not really concerned.

rmosolgo:
Double-checked this on a local project. Works & reloads on changes :)

rearadmiral:
I'm happy to see this merged. This is the key foundational step for a future flux-compatible plugin gem.

@rmosolgo did you ever have a chance to work on extension points for the base class? See my [comment above](#issuecomment-97610816) If you haven't spent any time on it, I can try to write one and see what everyone thinks

Thanks again for pushing this through.

rmosolgo:
Shoot, I thought I did that but obviously not :S I'm definitely open to it
though!

mchristen:
This seems to have introduced a regression when trying to use prerender:true and already stringified properties being passed into react_component.

For example, if you use jbuilder to build your JSON you could in the past(without this pull request) do something like

    react_component("MyComponent", render(file: "path/to/some_file.json.jbuilder), prerender: true)

It seems that the changes in view_helper.rb are the culprit. [view_helper.rb#L12](https://github.com/reactjs/react-rails/blob/981e3d9c40fa5f67ad715f94734dc522051225da/lib/react/rails/view_helper.rb#L12) assumes that the props aren't already stringified.

rmosolgo:
@mchristen thanks for reporting and finding the problem! I added some tests & worked up a fix here: #268

Guardfile
```diff
@@ -0,0 +1,6 @@
+guard :minitest do
```
vipulnsward: guard throws a circular loading in progress warning

PR #257, opened 2015-05-04
Sprockets cache expiration based on react-rails settings

byroot:
Followup to https://github.com/reactjs/react-rails/pull/254

See https://github.com/reactjs/react-rails/pull/254#issuecomment-98846865

cc  rmosolgo

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -41,6 +41,14 @@ module React
       end
 
       config.after_initialize do |app|
+        variant = app.config.react.variant == :production ? 'production' : 'development'
+        variant += '-with-addons' if app.config.react.addons
+
+        app.assets.version = [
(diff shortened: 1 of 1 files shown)


byroot:
I rebased on top of master now that #254 is merged. But it seems tests are failing. I'll find some time to fix them.

rmosolgo:
Is `version` relatively new? It's throwing a `NoMethodError` so maybe it's not defined in some of the older Rails versions we support (3.2+)

byroot:
No the problem is that `app.assets` is nil. Probably because the code I added is too soon in the initialization chain.

byroot:
I just moved that code on after_initialize. It should go green now. I just keep forgetting to use `appraisal` to test my changes.

rmosolgo:
:+1: thanks again for upstreaming!

PR #260, opened 2015-05-07
Use standard shortcode for Apache-2.0

mkenyon:
Hi!

Just a quick change to make license-finding tools like [license_finder][0] happier based off [industry standards][1].

Thanks!

[0]: https://github.com/pivotal/LicenseFinder
[1]: https://spdx.org/licenses/


zpao:
Yea this is fine, I meant to do that a long time ago.

We should probably actually switch to BSD+PATENTS as we've done with our other FB projects.

zpao:
This is a super trivial change so is fine to go in w/o CLA (though we'd love to see you contribute more, which will require it to be signed)

PR #262, opened 2015-05-11
Fixed bug with generating unusable components

### LINKED ISSUE #261: rails generate component builds an unusable component
Hi there,

First of all thanks for this great lib!

The bug I encountered is neither critical, nor hard to fix:
If you type `rails g react:component SubscriptionForm subscribable_type:string subscribable_id:number` you get a component which looks like this:

```
var SubscriptionForm = React.createClass({
  propTypes: {
    subscribableType: React.PropTypes.string,
    subscribableId: React.PropTypes.number
  },

  render: function() {
    return (
      <div>
        <div>Subscribable Type: {this.props.subscribable_type}</div>
        <div>Subscribable: {this.props.subscribable_id}</div>
      </div>
    );
  }
});
```

So the either the propTypes or the props used in the render function are correct. If I find time later I will maybe add a PR. 

Regards!

DanielMSchmidt:
Fixes #261

### DIFF
--- a/lib/generators/templates/component.js.jsx
+++ b/lib/generators/templates/component.js.jsx
@@ -12,7 +12,7 @@ var <%= file_name.camelize %> = React.createClass({
     return (
       <div>
 <% attributes.each do |attribute| -%>
-        <div><%= attribute[:name].titleize %>: {this.props.<%= attribute[:name] %>}</div>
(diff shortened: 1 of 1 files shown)


rmosolgo:
Good find, thanks for pushing the fix!

Issue #265, opened 2015-05-15
Namespacing JS Components and Server Side Rendering


jhubert:
I'm having a difficult time migrating over to server side rendering. My components are namespaced (MyApp.ComponentName) and the react_component helper / execjs keeps throwing Prerender errors:

```
Encountered error "ReferenceError: Can't find variable: MyApp" when prerendering MyApp.DashboardButtons with {}
```

My server_renderer_options are setup like this:
```ruby
config.react.server_renderer_options = {
  files: ["server_render.js"] # files to load for prerendering
}
```

(I've also tried `files: ["react.js", "init.js", "components.js"]`)

`server_render.js` looks like this:

```
//= require react
//= require react_ujs
//= require init
//= require components
```

`init.js` looks like this:
```javascript
window.MyApp = {};
window.MyApp.Mixins = {};
window.MyApp.util = {};
```

(I've also tried `self.`, `global.`, `this.`, etc.)

the component coffee file looks like this:

```coffeescript
MyApp.DashboardButtons = React.createClass
  displayName: 'DashboardButtons'

  render: ->
    R.div {}, 'Hello'
```

I'm calling the helper like this:  `react_component('MyApp.DashboardButtons', {}, {prerender: true})`

If I take out the namespace and change it to just `@DashboardButtons` in the coffee file, it works fine but the rest of my app (which uses the namespace) breaks.

How do I define the namespace so that it's used properly by ExecJS and my components are rendered properly? Any suggestions would be really appreciated.

rmosolgo:
What's your react-rails version? The README is correct for `master` , but the `1.0.0` server rendering options are different. 

See either the [1.0 readme](https://github.com/reactjs/react-rails/tree/v1.0.0#server-rendering) or run master from github: 

```ruby
gem 'react-rails',
  github: 'reactjs/react-rails',
  branch: 'master'
```

I'm sorry, I forgot this would make the readme out of date, I'm about to add a note about the API change!

### replicate? 

I tried to replicate this situation locally by setting up a `server_render.js` like:

```js
//= require react
//= require init
//= require components
```

and I was able to use components in a `MyApp` namespace.

rmosolgo:
By the way thanks for your very detailed report! :confetti_ball:

jhubert:
I used `gem 'react-rails', '~> 1.0'` to add react-rails, based on the current README in `master`. Whoops.

[...]

PR #266, opened 2015-05-15
fix(readme) add note about API version

rmosolgo:
Oops, forgot about the fact that my changes to server rendering would make the home page out of sync with rubygems. I think that might be what's happening in #265.

Soon I'll push `1.1` but I think this note will help for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ react_component(component_class_name, props={}, html_options={})
 
 ### Server rendering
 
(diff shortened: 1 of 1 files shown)


jhubert:
You may consider changing the `gem 'react-rails', '~> 1.0'` line in the README to 

```
gem 'react-rails',
  github: 'reactjs/react-rails',
  branch: 'master'
```

Until they are back in sync. I think that's where my confusion was coming from as well. I didn't realize that I wasn't using the version that matched the readme since I was following the readme to set it up.

Issue #249, opened 2015-04-23
Triggering Sprockets on Rails init


toolmantim:
Whilst trying to reduce our app boot time and memory I noticed Sprockets extension name parsing was being triggered on Rails app load.

```
$ ruby -rmemory_profiler -I. -e 'MemoryProfiler.report { require "config/environment" }.pretty_print'
Total allocated 2057572
Total retained 390073

allocated memory by gem
-----------------------------------
activesupport-4.2.1 x 60149050 (note: this is just because dependency require'ing gets reported under here)
sprockets-3.0.1 x 56626099
2.2.0/lib x 28229776
rubygems x 13925583

retained memory by gem
-----------------------------------
sprockets-3.0.1 x 14024689
2.2.0/lib x 7370079
activesupport-4.2.1 x 5285833
```

I added a `raise` inside `Sprockets#extname_map` that is causing all this, and the only place it's called during Rails init is from this part of railtie.rb:

```
      config.react.react_js = lambda {File.read(::Rails.application.assets.resolve('react.js'))}
```

Removing `react-rails` from the Gemfile brings things back to normal (`extname` is never called):

```
$ ruby -rmemory_profiler -I. -e 'MemoryProfiler.report { require "config/environment" }.pretty_print'
Total allocated 1614438
Total retained 257955

allocated memory by gem
-----------------------------------
activesupport-4.2.1 x 60005108
2.2.0/lib x 28192938
rubygems x 13108443
json-1.8.2 x 9075397
mime-types-2.4.3 x 7098487
erubis-2.7.0 x 5427205
```

So `config.react.react_js` is called from in the `config.after_initialize` like so:

```
          React::Renderer.setup!( cfg.react_js, cfg.components_js, cfg.replay_console,
                                {:size => cfg.max_renderers, :timeout => cfg.timeout})
```

Is there a way to avoid this call to `::Rails.application.assets.resolve('react.js')` in `cfg.react_js`?

mchristen:
This causes extreme app bootup times.

When hosted on Heroku which has a 60s timeout for app boot you can get yourself into some trouble after a server reboot where your servers are marked as crashed and no requests are processed. 

I find the app boot time to be right around 60s and it basically makes deploying to heroku a complete crapshoot on whether or not your app will boot back up.

toolmantim:
We've just removed this gem from our app altogether…

rmosolgo:
Thanks for your detailed report & info about how to measure the issue. This is new turf for me!

[...]

PR #267, opened 2015-05-15
Set the assets version from an initializer

byroot:
Possible fix for https://github.com/reactjs/react-rails/issues/249#issuecomment-102529896


rmosolgo:
:+1: fixed the reported issue!

PR #268, opened 2015-05-16
fix(prerender) accept string props

rmosolgo:
My first API was shoving `prerender:` into the props hash, but obviously that doesn't work if props isn't a hash! 

So instead, pass `prerender_options` along through the ServerRendering pipeline. In theory, a render could take any options. 

`SprocketsRenderer` looks for `:static`, to use `React.renderToStaticMarkup`

### DIFF
--- a/lib/react/rails/view_helper.rb
+++ b/lib/react/rails/view_helper.rb
@@ -9,7 +9,7 @@ module React
 
         prerender_options = options[:prerender]
         if prerender_options
-          block = Proc.new{ concat React::ServerRendering.render(name, props.merge(prerender: prerender_options)) }
+          block = Proc.new{ concat React::ServerRendering.render(name, props, prerender_options) }
         end
 
         html_options = options.reverse_merge(:data => {})
--- a/lib/react/server_rendering.rb
+++ b/lib/react/server_rendering.rb
@@ -11,9 +11,9 @@ module React
       @@pool = ConnectionPool.new(options) { create_renderer }
     end
 
-    def self.render(component_name, props)
+    def self.render(component_name, props, prerender_options)
       @@pool.with do |renderer|
-        renderer.render(component_name, props)
+        renderer.render(component_name, props, prerender_options)
       end
     end
 
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -14,13 +14,13 @@ module React
         @context = ExecJS.compile(js_code)
       end
 
-      def render(component_name, props)
+      def render(component_name, props, prerender_options)
         # pass prerender: :static to use renderToStaticMarkup
-        if props.is_a?(Hash) && props[:prerender] == :static
-          react_render_method = "renderToStaticMarkup"
-        else
-          react_render_method = "renderToString"
-        end
+        react_render_method = if prerender_options == :static
+            "renderToStaticMarkup"
+          else
+            "renderToString"
+          end
 
         if !props.is_a?(String)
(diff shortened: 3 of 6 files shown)


mchristen:
I tried this out and I'm running into an issue with sprockets now.

In particular 

    Asset was linked to from an alias rather than its exact path. Alias resolving may not be available in production.
    Use "react.js" instead of "development-with-addons/react.js"

being thrown from

    #<Class:0x007fee7182f188>#asset_digest_path
    sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb, line 134

This is kind of confusing because it is a call to //= require react that is raising the exception

rmosolgo:
wow, is there any more stack trace to that? 

If it's booting a renderer and _trying_ to prerender, I guess it's coming from here: https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/sprockets_renderer.rb#L11

But why would it work fine in the asset pipeline but not in prerender?

rmosolgo:
Reproduced on Sprockets 2.8. Oh, I misunderstood: it's coming from the asset pipeline, _not_ from prerender!

mchristen:
Correct, this is for the client side JS file loaded in the rails layout. 

     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:134:in `asset_digest_path'
     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:85:in `compute_asset_path'
     actionview (4.2.0) lib/action_view/helpers/asset_url_helper.rb:135:in `asset_path'  
     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:100:in `asset_path'
     actionview (4.2.0) lib/action_view/helpers/asset_url_helper.rb:245:in `javascript_path'
     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:153:in `block (2 levels) in javascript_include_tag'
     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:152:in `block in javascript_include_tag'
     sprockets-rails (2.3.1) lib/sprockets/rails/helper.rb:149:in `javascript_include_tag'
     app/views/layouts/application.html.haml:33:in `_app_views_layouts_application_html_haml___1984877869646349502_70331041908800'

rmosolgo:
Looks like this was touched recently in `sprockets-rails`, double-checking here: https://github.com/rails/sprockets-rails/issues/247#issuecomment-103162751

I didn't notice it before because I was on sprockets 3 :S

rmosolgo:
@mchristen sounds like a sprockets-rails 2.3.1 issue, see this comment: https://github.com/rails/sprockets-rails/issues/247#issuecomment-103666873

mchristen:
It definitely occurs prior to that version as well(at least as early as 2.2.4). When I first ran into the error I came across that issue on sprocket-rails and saw that code was touched with 2.3.1 so I updated to that version to see if the problem went away. When the problem persisted is when I made the comment above. 

Either way it looks like there will be a fix shortly, thanks for the work!

rmosolgo:
aiiiyayay well the regression test for this particular issue passes, so i'm gonna go for it

rmosolgo:
@mchristen they've shipped `sprockets-rails` 2.3.2 & I'm able to use them together:

```
~/code/check-ins $ bundle show | grep 'sprockets\|react'
  * react-rails (1.0.0 9a4f6ab)
  * sprockets (2.12.3)
  * sprockets-rails (2.3.2)
```

Please let me know if you can confirm/deny, if we're clear with the latest sprockets-rails I want to get ready for a new version release :D

PR #270, opened 2015-05-17
Fix Appraisals and test against the intended version of Rails

benpickles:
The pessimistic constraint was mis-targeted - for instance `~> 4.1` was actually specifying `>= 4.1 && < 5.0` instead of the intended `>= 4.1 && < 4.2` and thus the rails-4.1 Appraisal was actually testing against Rails 4.2.

I also dropped support for Rails 3.1 as it doesn't have `watchable_files` (https://github.com/reactjs/react-rails/blob/46bb0b5709611a9824b3b0bedcd910f92b7aebd4/lib/react/rails/railtie.rb#L20) and therefore needs some work. Also, I assume nobody's actually using react-rails with Rails 3.1 as there'd probably be a bug report by now.


benpickles:
Not sure why Travis is failing (`undefined method api_only` doesn't seem to mean anything in react-rails or rails itself), possibly just need to restart the job.

rmosolgo:
Thanks for fixing this! I _thought_ I knew how `~>` worked but apparently not :P

LGTM once Travis is green

rmosolgo:
Aaaand as you suspected, green after re-run!

## Before 

Only 2 Rails versions because minor versions were skipped:





## After

rmosolgo:
Thanks! :confetti_ball:

PR #273, opened 2015-05-21
refactor(JSX::Transformer) use Transformer class; allow asset_path config

rmosolgo:
Simplifying the `JSX` module and providing the feature from #239 

- Use an instance of `JSX::Transformer` for transforming JSX
- `JSX.transform_options` are initialization options for `Transformer`
- feature: allow custom JSX files by passing `asset_path:` config (or even overriding `Transformer#jsx_transform_code` :grimacing: )


rmosolgo:
I added that `clear_sprockets_cache` helper to the tests, one of these days I'll go through and weed out all the different ways that the cache is cleared and replace them with that helper

PR #274, opened 2015-05-24
Codeclimate coverage

rmosolgo:
picking up work from #142


rmosolgo:
@zpao can you help me out? I think I need the `CODECLIMATE_REPO_TOKEN`, which is under "Settings" > "Test Coverage". Can you see the gear icon here?

image

zpao:
Hmm, that's already set on Travis so I think you probably just need to push this (might only check for coverage on master).

rmosolgo:
So I see! Thanks, I looked in the repo and didn't find it, but it must be there:

image

PR #276, opened 2015-05-27
add a changelog

rmosolgo:
Oops, I knew something was missing. 

I've modeled this one after the React.js changelog.


danott:
:+1: This will be super helpful!

PR #280, opened 2015-05-27
Add `tilt` dependency

gbanis:
Same issue as #243 and #248.

This is a greater issue than just Rspec failing. rails g scaffold and several other commands that touch the dependency fail.


rmosolgo:
thanks for porting this change!

PR #286, opened 2015-05-29
This changes "::" in the component name in the react_component helper to "."

catmando:
So that opal react components read nicely when using react-rails

I did not see how to run the tests but I added a test hopefully

### DIFF
--- a/lib/react/rails/view_helper.rb
+++ b/lib/react/rails/view_helper.rb
@@ -5,6 +5,9 @@ module React
       # are used by react_ujs to actually instantiate the React component
       # on the client.
       def react_component(name, props = {}, options = {}, &block)
+        
+        name = name.gsub("::",".") # allows ruby style naming
+        
         options = {:tag => options} if options.is_a?(Symbol)
 
         prerender_options = options[:prerender]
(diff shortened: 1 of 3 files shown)


rmosolgo:
I could be underestimating the desire for this, but I think this is a very specific need. Are there others who want to do this?

If you wanted to use a different scheme for accessing components in your app, you could define your own view helper, for example:

```ruby
module ApplicationHelper
  def opal_react_component(module_style_name, props, render_options={})
    js_name = module_style_name.gsub("::", ".") # or any other transforms
    react_component(js_name, props, render_options)
  end
end
```

I'm happy with the existing features of this gem (mounting components by their JS identifiers) but if this is widely requested, I could be convinced!

catmando:
Thanks for the feedback!  

We are working hard on the react.rb gem that gives a full ruby wrapper to react (using the opal transpiler)  So the hope would be YES there will be LOTS of people wanting to do this.

That SAID your approach is much better.  I should be able include your suggestion into the react.rb gem with some conditional logic that checks to see if react-rails is being used.    I don't want react.rb to depend on react-rails, but likewise you should not have dependencies in react-rails!

Thanks again!

rmosolgo:
yes, thanks for the suggestion but I think it's a better fit for an Opal-specific gem!

PR #293, opened 2015-06-10
#WIP - attempt to parse Slim instead of Html in jsx

reneweteling:
* [ ] Listen to a new extension handler if slim is used probably .js.jsx.slim.coffee
* [ ] Parse the slim into html
* [ ] Take care of the class attribute --> className attribute

### DIFF
--- a/lib/react/jsx.rb
+++ b/lib/react/jsx.rb
@@ -14,6 +14,14 @@ module React
     self.transformer_class = Transformer
 
     def self.transform(code)
+      # slim = code.match( /<slim>(.*)<\/slim>/m )
+
+      code = code.gsub( /<slim>(.*)<\/slim>/m ) {|slim| 
+        Slim::Template.new { slim }.render
+      }
+
+      
+      
       transformer.transform(code)
     end
 


rmosolgo:
Wow, this would be quite a feat! However I think a dependency on Slim is outside the scope of react-rails. 

The master branch has a nice extensibility point for this, though. You could define your own transformer class add assign it to `JSX.transformer_class`: https://github.com/reactjs/react-rails/blob/master/lib/react/jsx.rb#L10-L14

What do you think of that? 

If you end up making it work, I'd be happy to add an "add-ons" page to the Wiki & include this project as the first entry :)

reneweteling:
@rmosolgo nice one! took me a while before i noticed its not in the gem but on the master :) but im playnig with it thanks for the heads up

rmosolgo:
Yeah, hoping to cut a new version soon, but waiting for https://github.com/rails/sprockets-rails/issues/257

rmosolgo:
Please do let me/us know if it comes together, it would be great to add a reference on this project somewhere!

Issue #278, opened 2015-05-27
ES6 not compiling


desmondhume:
I'm trying to use ES6 features in this react component:

```jsx
import React from "react/addons";
import {GoogleMaps} from "react-google-maps";

class SimpleMap extends React.Component {
  render () {
    const {props, state} = this,
          {googleMapsApi} = props;
    return (
      <GoogleMaps containerProps={{
          style: {
            height: "100%",
          },
        }}
        googleMapsApi={google.maps}
        zoom={8}
        center={{lat: -34.397, lng: 150.644}} />
    );
  }

}

export default SimpleMap;
```

but I keep getting this error no matter what i do:

```
Error: Parse Error: Line 2: Illegal import declaration
  (in `[...]app/assets/javascripts/components/simple_map.js.jsx`)
```

I'm using react-rails along with [browserify-rails](https://github.com/browserify-rails/browserify-rails).

In my `config/application.rb` I put: 
```
config.react.jsx_transform_options = {
  harmony: true
}
```

I don't know what to try next...any suggestion/idea?

krzysiek1507:
Do you try without <code>import</code>?

rmosolgo:
the `harmony` option currently supports _some_ ES6 features, but not all. I've never found a clear-cut list of features it supports but you can see the visitors (which correspond to transforms) here:  https://github.com/facebook/jstransform/tree/master/visitors

tirdadc:
I was [trying to use computed property names](http://stackoverflow.com/a/30900321/3102766) and it seems to be [supported](https://github.com/facebook/jstransform/blob/master/visitors/es6-object-computed-property-visitors.js), but despite setting the harmony flag to true I get `Uncaught SyntaxError: Unexpected token [` when it hits the new syntax.

vipulnsward:
These will go away with https://github.com/reactjs/react-rails/pull/295

rmosolgo:
Try `1.1.0`, it uses @vipulnsward 's new BabelTransformer :smile:

PR #295, opened 2015-06-11
New Babel Transformer

### LINKED ISSUE #292: Use Babel to transform JS
We're going to deprecate JSXTransformer and probably not ship any more versions of it. We've already switched the React build process to use Babel and we're going to encourage others to as well. It would make sense for react-rails to follow suit. It would also fix issues like #278.

The `babel-transpiler` gem looks like it should be pretty easy to drop in. https://github.com/babel/ruby-babel-transpiler

vipulnsward:
- Added new BabelTransformer to start performing transformation using ruby-babel-transpiler
- Make new BabelTransformer the default transformer
- README changes to reflect new Babel Transformer
- Added deprecation warning for usage of old transformer options
- Move old transformer to JSXTransformer
- Fixed  component generator test after changes to new transformer

Fixes #292, #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,16 +114,6 @@ config.react.jsx_transform_options = {
 }
 ```
 
-#### CoffeeScript
-
-To use CoffeeScript, create `.js.jsx.coffee` files and embed JSX inside backticks, for example:
-
-```coffee
-Component = React.createClass
-  render: ->
-    `<ExampleComponent videos={this.props.videos} />`
-```
-
 ### Rendering & mounting
 
 `react-rails` includes a view helper (`react_component`) and an unobtrusive JavaScript driver (`react_ujs`) 
@@ -289,8 +279,9 @@ end
 
 ## CoffeeScript
 
-It is possible to use JSX with CoffeeScript. We need to embed JSX inside backticks 
-so CoffeeScript ignores the syntax it doesn't understand. Here's an example:
+It is possible to use JSX with CoffeeScript. To use CoffeeScript, create files with an extension `.js.jsx.coffee`. 
+We also need to embed JSX code inside backticks so that CoffeeScript ignores the syntax it doesn't understand.
+Here's an example:
 
 ```coffee
 Component = React.createClass
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@
 # react-rails
 
 
-`react-rails` makes it easy to use [React](http://facebook.github.io/react/) and [JSX](http://facebook.github.io/react/docs/jsx-in-depth.html) in your Ruby on Rails (3.2+) application. `react-rails` can:
+`react-rails` makes it easy to use [React](http://facebook.github.io/react/) and [JSX](http://facebook.github.io/react/docs/jsx-in-depth.html) 
+in your Ruby on Rails (3.2+) application. `react-rails` can:
 
 - Provide [various `react` builds](#reactjs-builds) to your asset bundle
 - Transform [`.jsx` in the asset pipeline](#jsx)
@@ -32,7 +33,8 @@ rails g react:install
 ```
 
 This will:
-- create a `components.js` manifest file and a `app/assets/javascripts/components/` directory, where you will put your components
+- create a `components.js` manifest file and a `app/assets/javascripts/components/` directory, 
+where you will put your components
 - place the following in your `application.js`:
 
   ```js
@@ -45,7 +47,8 @@ This will:
 
 ### React.js builds
 
-You can pick which React.js build (development, production, with or without [add-ons]((http://facebook.github.io/react/docs/addons.html))) to serve in each environment by adding a config. Here are the defaults:
+You can pick which React.js build (development, production, with or without [add-ons]((http://facebook.github.io/react/docs/addons.html))) 
+to serve in each environment by adding a config. Here are the defaults:
 
 ```ruby
 # config/environments/development.rb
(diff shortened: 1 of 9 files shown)


vipulnsward:
Couple of notes about the default transformer options:
- I have referred https://github.com/facebook/react/blob/master/bin/jsx-internal to set these options
- plugins: [constants] is not part of the options right now since we don't have it in the sources, should this be a part in future?

rmosolgo:
:clap: :clap: Your implementation looks great!

I don't understand your question about "plugins: [constants]" (I'm not familiar with babel yet), what do you mean by that?

README.md
```diff
@@ -85,6 +114,8 @@ config.react.jsx_transform_options = {
 }
 ```
 
+#### CoffeeScript
```
rmosolgo: Should we merge this with the "## Coffeescript" section below?

lib/react/jsx/babel_transformer.rb
```diff
@@ -0,0 +1,29 @@
+require 'babel/transpiler'
+module React
+  module JSX
+    class BabelTransformer
+      DEPRECATED_OPTIONS = [:harmony, :strip_types, :asset_path]
+
+      def initialize(options)
+        _options = options.dup
+        has_old_opts = DEPRECATED_OPTIONS.map do |option|
```
rmosolgo: Is there any downside to passing the deprecated keys to Babel? Are they ignored or do they raise errors?

If they're just ignored, this could be simplified a bit:

```ruby
if (options.keys & DEPRECATED_OPTIONS).any?
  # deprecation warning ...
end
```

lib/react/jsx/babel_transformer.rb
```diff
@@ -0,0 +1,29 @@
...
+          _options.delete option
+        end.any?
+
+        if has_old_opts
+          ActiveSupport::Deprecation.warn("Setting config.react.jsx_transform_options for :harmony, :strip_types, and :asset_path keys is now deprecated and has no effect with the default Babel Transformer."+
+                                              "Please use new Babel Transformer options :whitelist, :plugin instead.")
+        end
+
+        @transform_options = {
```
rmosolgo: How about moving the defaults to `DEFAULT_OPTIONS` ?

Then just 

```ruby
@transform_options = DEFAULT_OPTIONS.merge(options)
```

lib/react/jsx/babel_transformer.rb
```diff
@@ -0,0 +1,29 @@
...
+            blacklist: ['spec.functionName', 'validation.react']
+        }.merge(_options)
+      end
+
+      def transform(code)
+        puts @transform_options
```
rmosolgo: oops, probably don't need this anymore :)

rmosolgo:
Could we get a few minimal tests for `BabelTransformer` ? IMO it doesn't have to be much, but it would be nice to have a test to ensure some future change doesn't totally break it!

netikular:
Sorry to just jump in here, but does this implementation by any chance have any relation to the work done to support es6 in sprockets?

https://github.com/TannerRogalsky/sprockets-es6

Thanks for the great work!

rmosolgo:
@netikular not at all, "just jumping in" is the beauty of a public project, thanks for speaking up :)

I noticed that project as I was sizing up some options. I didn't it very compelling because: 

- the documentation was quite sparse (how do you pass Babel options? how could we make it process `.jsx` files?)
- the readme wasn't compelling ("This plugin is primarily experimental and will never reach a stable 1.0.")

So, I think it's ok to have a custom transformer!

vipulnsward:
I have been using this alongside react-rails in our application without any issues, to support es6.
Right now its on track to be merged in sprockets-rails. 
I think we can start looking at handling it, after it lands in sprockets-rails master.

netikular:
@rmosolgo I agree that the documentation is quite sparse, it took me a little while to get it working.

@vipulnsward are you using es6 and JSX together or are you writing your react components in the react-rails style and using the helpers?  I noticed that I can rename my JSX files to es6 and with some hoisting up to window get them to work, but it seems like a lot of setup work.

I was just wondering how you are handling working between es6 and jsx and if you are using a module loader or just es6 syntax.

rmosolgo:
:+1: if you guys prefer it, I'm open to it!

vipulnsward:
Ok, here's the current status- 
- One test fails related to coffeescript due to http://babeljs.io/docs/faq/#why-is-this-being-remapped-to-undefined- . 
Babel currently has strict mode enabled by default, hence the breaking of tests. It can be disabled(not recommended), or we would need to find workaround only for coffeescript files.
- plugins: [constants] - this option is defined here - https://github.com/facebook/react/blob/master/bin/jsx-internal#L41 and is a plugin loaded from https://github.com/facebook/react/blob/master/vendor/constants.js and is not included here.  This is a question for @zpao I guess, about whether its needed here, I don't think it applies here.
- Test taken care of

zpao:
The plugins: constants bit in the React build config shouldn't apply here. That's a custom plugin to transform `__DEV__` away. We never shipped that with JSXTransformer.

I don't know how custom plugins would hook into this, I'll let you figure that out :) Might be best to not support them initially unless you've tested and documented how it works.

rmosolgo:
Cool, sounds like we don't need `plugins: [constants]`. I agree, custom plugins can be a "later" thing.

As for coffee-script, what if we left `JSXTransformer` as the default for `1.1`, then switch the default and add a warning for coffee-script users in `1.2`? That keeps things working for people now & keeps our project moving forward.

Either way, I don't think it's a deal-breaker. `.coffee` users have two decent options: use `JSXTransformer` or disable strict mode in babel. (And a third very good option: switch to JSX :P)

vipulnsward:
Custom plugins are as simple as just passing to `plugin` transform_options. 

I think instead of shipping with `JSXTransformer` as default, we can ship with Babel as default(to be in sync with React) with disabled `strict` mode to support both coffeescript + JSX and make it explicit that next version default transformer will break CS. Thoughts?

PS: Even if we ship with breaking changes for CS later, Babel can still be used with passing custom transform options - `{ blacklist: ['spec.functionName', 'validation.react', 'strict'] }`

vipulnsward:
I have pushed an updated version with above changes. I think it makes sense to disable strict by default, to be backwards compatible, since thats how `JSXTransformer` is handling it.

vipulnsward:
Travis looks happy :-)

rmosolgo:
:+1: looks good to me :grinning:

rmosolgo:
:confetti_ball: the future is now :smile:

PR #297, opened 2015-06-15
Make indentation in heredocs consistent

glittershark:
Always indent one space inside <<-JS heredocs, rather than indenting in
some places and not in others


rmosolgo:
:+1:

PR #290, opened 2015-06-08
[testing] Execjs runtime benchmarks

rmosolgo:
(I don't really want to merge this, I just want to share some code & results) 

I ran some tests to compare ExecJS runtimes & connection pool sizes:

- I added `ExecJSRenderer` to sidestep the dependency on Rails asset pipeline
- I threw [duktape.rb](https://github.com/judofyr/duktape.rb) in the mix just for kicks
- update: try benchmark with lots of threads accessing the lock

## Results 

```
rmosolgo ~/code/react-rails $ ruby -I lib  benchmarks/server_rendering_benchmark.rb
                                          user     system      total        real
threaded, 1 conn, therubyracer (V8)   0.120000   0.010000   0.130000 (  0.127382)
threaded, 1 conn, Duktape             0.460000   0.010000   0.470000 (  0.465932)
threaded, 1 conn, JavaScriptCore      0.110000   0.250000   2.460000 (  2.487619)
threaded, 1 conn, Node.js (V8)        0.110000   0.250000   9.560000 (  9.504323)
threaded, 10 conn, therubyracer (V8)  0.070000   0.010000   0.080000 (  0.074837)
threaded, 10 conn, Duktape            1.480000   0.040000   1.520000 (  1.515937)
threaded, 10 conn, JavaScriptCore     0.080000   0.230000   3.840000 (  0.932142)
threaded, 10 conn, Node.js (V8)       0.140000   0.250000  16.970000 (  3.768226)
threaded, 25 conn, therubyracer (V8)  0.060000   0.010000   0.070000 (  0.068196)
threaded, 25 conn, Duktape            3.170000   0.060000   3.230000 (  3.230653)
threaded, 25 conn, JavaScriptCore     0.190000   0.300000   4.240000 (  1.856800)
threaded, 25 conn, Node.js (V8)       0.080000   0.240000  17.570000 (  5.963759)
1 conn, therubyracer (V8)             0.070000   0.010000   0.080000 (  0.080440)
1 conn, Duktape                       0.470000   0.020000   0.490000 (  0.480219)
1 conn, JavaScriptCore                0.090000   0.230000   2.400000 (  2.442591)
1 conn, Node.js (V8)                  0.100000   0.250000   9.360000 (  9.294853)
10 conn, therubyracer (V8)            0.060000   0.010000   0.070000 (  0.079364)
10 conn, Duktape                      1.530000   0.040000   1.570000 (  1.558580)
[...]

PR #299, opened 2015-06-15
feat(ExecJSRenderer) add ExecJSRenderer

rmosolgo:
This `ExecJSRenderer` was created while [testing server rendering performance](https://github.com/reactjs/react-rails/pull/290) and I thought it might be a chance to reorganize server rendering code. In this branch: 

- __`ExecJSRenderer`__ is basically a wrapper around ExecJS
- __`SprocketsRenderer`__ is an `ExecJSRenderer` which fetches source JS from the asset pipeline & enables "console replay"

What do you think?

## Pros: 

- ExecJSRenderer was great for testing performance. I'm not sure what other use it could have but it might come in handy again. 
- ExecJSRenderer's `#before_render` and `#after_render` hooks add extensibility as requested in https://github.com/reactjs/react-rails/pull/253#issuecomment-102515499
- It demonstrates how Renderers fit into `react-rails`, maybe people will make their own 

## Cons: 

- The split is a bit arbitrary, specifically:
  - Where should "console replay" be implemented?
  - Which renderer decides between `renderToString` and `renderToStaticMarkup` ? Right now, that's very awkward.
- ExecJSRenderer might be a pointless exercise because `react-rails` _depends_ on sprockets, why wouldn't you use `SprocketsRenderer`?

Anyways, curious what others think of this change!

### DIFF
--- a/lib/react/server_rendering.rb
+++ b/lib/react/server_rendering.rb
@@ -1,4 +1,5 @@
 require 'connection_pool'
+require 'react/server_rendering/exec_js_renderer'
 require 'react/server_rendering/sprockets_renderer'
 
 module React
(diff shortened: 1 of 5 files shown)


rmosolgo:
no news is good news

PR #302, opened 2015-06-23
Config & initialization fixes

rmosolgo:
- Accept transformer class from config
  - Right now the only way to do that was assign `React::JSX.transformer_class = ` which is ok with me, but I guess a Rails config is more normal 

- Don't reload the transformer in dev
  - Make development match production
  - If you change your JSX config, you'll need to restart the app anyways 

- Move JSXTransformerTest 
  - one class, one file: works nice with Guard, makes it easy to find the tests for a class
  - (I know not all of react-rails is on that pattern yet but i'm working on it :) 

- fix(ServerRendering) default pool_size of 1; document deadlock & ther… 
 - based on findings from #290  

- refactor(AssetVariant) extract asset build & path logic
  - just want to DRY it up and get logic out of the Railtie

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ You can configure your pool of JS virtual machines and specify where it should l
 # These are the defaults if you dont specify any yourself
 MyApp::Application.configure do
   # Settings for the pool of renderers:
-  config.react.server_renderer_pool_size  ||= 10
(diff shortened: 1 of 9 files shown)


danott:
Great work! :100:

PR #306, opened 2015-06-29
Inline props operations improvement

panasyuk:
# Problem
1. Size of inline props placed to the root component tag as attributes may be too huge to:
  1.1 Process in browser devtools. So that browser stucks trying to render markup for preview.
  1.2 Deliver it to user due to bandwidth (for highload apps).
2. Inline props placed to the attributes of tag are looking freaky because of its size and HTML safety.

So props we want to place:
```JSON
{"foo": "bar"}
```
are looking like this:
```HTML
{&quot;foo&quot;:&quot;bar&quot;}
```

# Solution
Moving inline props to the ```<script type="text/json">{"foo": "bar"}</script>``` will:
1. Reduce the size of data passed to the page in order to render component at the initial state. It's also useful because it reduces bandwidth utilization (even if we are using GZIP).
2. Allow browser inspector to operate on DOM faster.
3. Make markup looking more cleaner.

Placing inline props at the end of body speeds up DOM rendering and allows Turbolinks to load this data along with other markup on each request.

# How it works
The process of placing the components to the markup in Rails is still the same except of developer should add ```<%= render_react_props %>``` at the end of ```<body>``` tag.

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -10,7 +10,6 @@
     CLASS_NAME_ATTR: 'data-react-class',
     PROPS_ATTR: 'data-react-props',
     PROPS_ID_ATTR: 'data-react-props-id',
-    SEPARATE_PROPS: <%= Rails.configuration.react.separate_props %>,
     RAILS_ENV_DEVELOPMENT: <%= Rails.env == "development" %>,
     // helper method for the mount and unmount methods to find the
     // `data-react-class` DOM elements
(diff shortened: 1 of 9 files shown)


rmosolgo:
Hi thanks for updating this, I'm sorry it's sat cold for quite some time! 

I hope to add an extension point for the view helper much like we have for transforming JSX & server rendering. Then you could define some class with the inspected behavior and inject it during configuration. That allows us to have lots of different options without packing too much behavior into one class or method. 

Thanks again for your work on identifying this problem & proposing a solution, I'll follow up soon!

rmosolgo:
Hoping to support more customization of the view helper with #346 !

That way you could distribute a custom helper & renderer in a gem and hook them up with a railtie.

panasyuk:
I would like present a GEM for this PR:
https://github.com/panasyuk/react-rails-view_helper_implementation

rmosolgo:
Nice! Glad that worked out.

PR #308, opened 2015-06-30
update therubyracer for tests

rmosolgo:
Ug, somehow I caused the test suite to fail in #302.

- 1.9.3 & JRuby fail on a `%i(...)` (but why now and not before?)
- Others fail on a `Segmentation fault` during JS execution :weary:


rmosolgo:
aaaand re-running `master` just worked :S

Issue #296, opened 2015-06-15
Why unmount on page change?


markbrown4:
"On page load, the react_ujs driver will scan the page and mount components using data-react-class and data-react-props. Before page unload, it will unmount components (if you want to disable this behavior, remove data-react-class attribute in componentDidMount)."

Why is this the default?  Removing chunks of the DOM when changing pages seems strange.

```
componentDidMount: ->
  @getDOMNode().parentNode.removeAttribute('data-react-class')
```

danott:
Someone else can correct me if I'm wrong, but I understood that the motivation is so that multiple calls to `ReactRailsUJS.mountComponents` don't clobber an already mounted component. If, for instance, an `ajax` call adds some content to the page, and a call is made to `ReactRailsUJS.mountComponents`, the minimal number of `React.render` calls are needed to get the DOM to the desired state.

markbrown4:
My question was regarding the last part of that sentence to do with the unmounting on page unload.

A full page change always blows away the DOM and starts again, can't think of a valid reason for doing extra work and causing re-rendering / painting before navigating.  For things like turbolinks this may make sense but there's special events you can use in those cases.

rmosolgo:
For me, it just seems like the "right" thing to do: React components have lifecycle hooks and it seems good to run through them. If people _want_ to skip those lifecycle hooks, ok, go for it, but executing them seems like a good default behavior. 

I imagine another simple work around would be:

```
ReactRailsUJS.unmountComponents = function() {}
```

Do you know how any other React + server framework libraries handle it? Do they unmount on unload?

markbrown4:
Disagree with you there, it's the wrong thing to do on page unload.  The only time it makes sense is for single page apps.

The way it behaves currently is really strange from a ux perspective and just unnecessary work for the browser slowing down page changes.

rmosolgo:
I definitely see your point, I'm open to a PR with a better handling of this case!

PR #318, opened 2015-07-09
remove unmountComponents on window unload

markbrown4:
Removing calls to ReactRailsUJS.unmountComponents on window unload for traditional apps using full page refreshes.  Unmounting components is unnecessary work on unload and causes undesirable ux when the DOM elements are removed causing repaints.  #296


rmosolgo:
I'm game for this but I wonder, does anyone see any red flags to _not unmounting_ when a user closes a page? 

Does anyone depend on this behavior?

rmosolgo:
:confetti_ball:  thanks!

PR #322, opened 2015-07-18
Fix running on sprockets-rails master

vipulnsward:
- Assets version and register_engine have been moved from app.assets to app.config.assets on sprockets
- Added a check to make sure to use proper asset object when calling version, register_engine, etc

### DIFF
--- a/lib/react/rails/engine.rb
+++ b/lib/react/rails/engine.rb
@@ -3,7 +3,7 @@ module React
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
         if app.assets.nil?
-         app.config.assets.register_engine '.jsx', React::JSX::Template
+          Sprockets.register_engine  '.jsx', React::JSX::Template
         else
           app.assets.register_engine '.jsx', React::JSX::Template
         end
--- a/lib/react/rails/engine.rb
+++ b/lib/react/rails/engine.rb
@@ -2,7 +2,11 @@ module React
   module Rails
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
-        app.assets.register_engine '.jsx', React::JSX::Template
+        if app.assets.nil?
+         app.config.assets.register_engine '.jsx', React::JSX::Template
+        else
+          app.assets.register_engine '.jsx', React::JSX::Template
+        end
       end
     end
   end
--- a/lib/react/rails/engine.rb
+++ b/lib/react/rails/engine.rb
@@ -2,11 +2,8 @@ module React
   module Rails
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
-        if app.assets.nil?
-          Sprockets.register_engine  '.jsx', React::JSX::Template
-        else
-          app.assets.register_engine '.jsx', React::JSX::Template
-        end
+        sprockets_env = app.assets || Sprockets # Sprockets 3.x expects this in a different place
+        sprockets_env.register_engine(".jsx", React::JSX::Template)
       end
     end
   end
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -38,10 +38,12 @@ module React
           addons: app.config.react.addons,
         })
 
-        app.assets.version = [
-          app.assets.version,
-          "react-#{asset_variant.react_build}",
-        ].compact.join('-')
+        if app.assets.nil?
+          app.config.assets.version = [app.config.assets.version, "react-#{asset_variant.react_build}",].compact.join('-')
+        else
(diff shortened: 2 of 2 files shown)


lib/react/rails/engine.rb
```diff
@@ -2,7 +2,11 @@ module React
   module Rails
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
-        app.assets.register_engine '.jsx', React::JSX::Template
+        if app.assets.nil?
+          Sprockets.register_engine  '.jsx', React::JSX::Template
+        else
+          app.assets.register_engine '.jsx', React::JSX::Template
+        end
```
krzysiek1507:
```ruby
(app.assets || Sprockets).register_engine '.jsx', React::JSX::Template```

Maybe this way?

Comment (vipulnsward) [inline lib/react/rails/engine.rb]: Actually its harder to read that way and make the intent clear why we are doing it.

**Comment by vipulnsward** [comment]
Hi @rmosolgo, can you take a look at this when you time. This causes issues when used with newest version of sprockets.

File: lib/react/rails/engine.rb (hunk 1 above)
Comment (rmosolgo) [inline lib/react/rails/engine.rb]: I also like the idea of removing that duplication, maybe:

```ruby 
sprockets_env = app.assets || Sprockets # Sprockets 3.x expects this in a different place 
sprockets_env.register_engine(".jsx", React::JSX::Template)
```

IMO less duplication = easier to maintain in the future!

lib/react/rails/railtie.rb
```diff
@@ -38,10 +38,12 @@ class Railtie < ::Rails::Railtie
...
         })
 
-        app.assets.version = [
-          app.assets.version,
-          "react-#{asset_variant.react_build}",
-        ].compact.join('-')
+        if app.assets.nil?
+          app.config.assets.version = [app.config.assets.version, "react-#{asset_variant.react_build}",].compact.join('-')
+        else
+          app.assets.version = [app.assets.version, "react-#{asset_variant.react_build}",].compact.join('-')
```
rmosolgo: I think this could also be DRYer, eg 

```ruby
sprockets_env = app.assets || app.config.assets # sprockets-rails 3.x attaches this at a different config
sprockets_env.version = [ ... ].compact.join("-")
```

lib/react/rails/engine.rb
krzysiek1507: +1

vipulnsward:
@rmosolgo done.

rmosolgo:
:confetti_ball:  nice! maybe we should add an Appraisal for Rails 5 beta or something next.

rovr:
@rmosolgo, @vipulnsward  Just a heads up, I've been trying to make it work with Rails 5 and atm it immediately fails on .register_engine line.

Looks like sprockets [removed engines completely](https://github.com/rails/sprockets/commit/37a1e24b0dafe0605ed0fc05e9a8c640caa21bb4).

So with sprockets 4.0 the initializer should *probably* look something like:

        sprockets_env.register_mime_type 'text/jsx', extensions: ['.jsx', '.js.jsx']
        sprockets_env.register_transformer 'text/jsx', 'application/javascript', React::JSX::Template

vipulnsward:
🙈 Sprockets is unstable atm with all the refactoring. I will work on
fixing this against master and keep it backwards compatible.

On Wednesday, September 2, 2015, rovr <notifications@github.com> wrote:

> @rmosolgo <https://github.com/rmosolgo>, @vipulnsward
> <https://github.com/vipulnsward> Just a heads up, I've been trying to
> make it work with Rails 5 and atm it immediately fails on .register_engine
> line.
>
> Looks like sprockets removed engines completely
> <https://github.com/rails/sprockets/commit/37a1e24b0dafe0605ed0fc05e9a8c640caa21bb4>
> .
>
> So with sprockets 4.0 the initializer should *probably* look something
> like:
>
>     sprockets_env.register_mime_type 'text/jsx', extensions: ['.jsx', '.js.jsx']
>     sprockets_env.register_transformer 'text/jsx', 'application/javascript', React::JSX::Template
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/reactjs/react-rails/pull/322#issuecomment-137001720>.
>


-- 
Vipul A.M.
+91-8149-204995

rovr:
@vipulnsward Thanks! I understand, I just thought I might save you some debugging time in the future.

So if sprockets choose to go with the current interface:

The object that you pass to the `register_transformer` method must implement a `call` method.

So if you do 

`sprockets_env.register_transformer 'text/jsx', 'application/javascript', React::JSX`

Then inside of React::JSX you must have a method like:

     def self.call(code)
       self.transformer ||= transformer_class.new(transform_options)
       self.transformer.transform(code[:data])
     end

Looks like that's all it takes to make it work with the current sprockets master branch.

jondot:
Bumping into this now. if this helps @rovr's fork works for me with current Rails master - thanks!

PR #324, opened 2015-07-23
Start testing against rails master

vipulnsward:
Should be green currently


rmosolgo:
woahhhhhh

> Rails 5 requires Ruby 2.2.2 or newer.

vipulnsward:
Hah, I forgot.

PR #329, opened 2015-07-31
Allow component inline rendering from controller

garbles:
Adds a `component` option to the controller `render` method so that you can render a component in place of `yield` instead of a view.

For example,

```ruby
class TodoController < ApplicationController
  def index
    render component: 'TodoList', props: { todos: [{...}, {...}] }
  end
end
```

I had done something similar to this this before [here](https://github.com/reactjs/react-rails/pull/118), but there were problems with it.


borisrorsvort:
:+1:

borisrorsvort:
@rmosolgo This is cool :) What is it missing to go on?

garbles:
@rmosolgo @zpao - wondering what else needs to be done here?

zpao:
I think it looks cool! This is @rmosolgo's show though so I'll let him take care of it.

lib/react/rails/controller_renderer.rb
```diff
@@ -0,0 +1,18 @@
+class React::ControllerRenderer
```
rmosolgo: Can this be named `React::Rails::ControllerRenderer`? That would match the file path (and seems like the right name for it)

test/server_rendered_html_test.rb
```diff
@@ -51,4 +51,9 @@ def wait_to_ensure_asset_pipeline_detects_changes
     assert_no_match(/console.warn/, response.body)
     assert_no_match(/console.error/, response.body)
   end
+
+  test 'react inline component rendering' do
+    get '/server/inline_component'
+    assert_match(/Render this inline/, response.body)
```
rmosolgo: What do you think about double-checking that this is rendered inside layout, eg testing for `<html>` and/or `<body>`?

rmosolgo: maybe also a test for the `tag:` option (checking for the span with `data-react-class` attribute)

rmosolgo:
Sorry I'm behind, thanks for the nudge! 

Wow, this implementation is really nice. I never knew about the `Renderers` API before. This seems like a great fit. 

I had a couple of questions I left in line. Besides that, could you add a bit to the readme?

borisrorsvort:
@rmosolgo  thanks for the review. Hope @garbles have time  to get this right :smile:  Can't wait :stuck_out_tongue:

test/server_rendered_html_test.rb
garbles: Added assertions for the tag, prerendering (`Render this inline</span>` instead of just `Render this inline`) and the layout (`<title>Dummy</title>`).

lib/react/rails/controller_renderer.rb
garbles: Whoops. Yup you're right. :+1:

garbles:
@rmosolgo @borisrorsvort I believe that the issues when the PR have been resolved. :family:

rmosolgo:
:money_with_wings: :confetti_ball:  thanks!

lib/react/rails/controller_renderer.rb
```diff
@@ -0,0 +1,17 @@
+class React::Rails::ControllerRenderer
+  include React::Rails::ViewHelper
+  include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::TextHelper
+
+  attr_accessor :output_buffer
```
rmosolgo: Btw @garbles these few calls (`include ActionView::...` + `attr_accessor :output_buffer`), is this what's required to make your custom object work with view rendering?

garbles: You are correct

garbles:
@rmosolgo Should I add something to the README?

rmosolgo:
Yeah, that would be good. Can you include a note like `(coming in 1.3.0)` or something? Some way that will signal to people that it's on master but not on Rubygems yet?

lib/react/rails/controller_renderer.rb
rmosolgo: cool, I was trying to refactor the view helper a while back and got scared off by how things kept _not working_, maybe i'll give it a try

garbles:
@rmosolgo cool cool. I'll write that up tonight. :tada:

borisrorsvort:
Giphy

Issue #316, opened 2015-07-09
Add es6 option to generator


borisrorsvort:
That could be a nice improvement to add an es6 version to the component generator :)

rstormsf:
agreed.

rmosolgo:
I'd support it!

borisrorsvort:
:+1:  @rmosolgo just to be clear, you'd support a PR or you'll include it in your roadmap?

rmosolgo:
if you make the PR, I'll support you :)

borisrorsvort:
@rmosolgo here you go :) https://github.com/reactjs/react-rails/pull/332

rmosolgo:
:confetti_ball:

PR #332, opened 2015-08-08
Add es6 option to generator

borisrorsvort:
Here you go #316

I wanted to add some tests but basically it's just rerunning the same with another option. Should I dedupe the current one just for es6 or something else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ end
 ### Component generator
 
 `react-rails` ships with a Rails generator to help you get started with a simple component scaffold. 
-You can run it using `rails generate react:component ComponentName`. 
+You can run it using `rails generate react:component ComponentName (--es6)`. 
 The generator takes an optional list of arguments for default propTypes, 
 which follow the conventions set in the [Reusable Components](http://facebook.github.io/react/docs/reusable-components.html) 
 section of the React documentation.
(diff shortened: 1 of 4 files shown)


rmosolgo:
Yes, could you add some tests for the generated output? That will help us keep from breaking things down the road :)

borisrorsvort:
@rmosolgo Good enough?

rmosolgo:
Sorry, I guess I should have been more specific. I think it's important to add tests for _new_ behaviors, but not important to test shared, pre-existing behaviors. In this case, it would be good to test: 

- The resulting file uses ES6 syntax (eg, checking for the `class ... extends` syntax in the result)
- The resulting file assigns propTypes after defining the class (eg, checking for `{className}.propTypes = `)
- The resulting file compiles successfully (I think `test "generates working jsx"` is sufficient for that). 

In my opinion, it's not important to repeat the tests for filename, `render` function, and `propType` since those behaviors are unchanged by the `--es6` option. 

How does that sound?

borisrorsvort:
@rmosolgo here you go

rmosolgo:
:money_with_wings:  Thanks, that's great!

borisrorsvort:
:+1:

PR #340, opened 2015-08-23
Allow to pass DOM object as parent selector.

aratak:
Extend possibility of `findDOMNodes` and as a consequence `mountComponents`. It allows to use it with an element which was inserted to the DOM tree without react.js. Example:

```
$(document).on('nested:fieldAdded', function(e) {
  window.ReactRailsUJS.mountComponents(e.target);
});
```

With backward compatibility, of course.

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -14,17 +14,29 @@
     // `data-react-class` DOM elements
     findDOMNodes: function(searchSelector) {
       // we will use fully qualified paths as we do not bind the callbacks
-      var selector;
-      if (typeof searchSelector === 'undefined') {
-        var selector = '[' + window.ReactRailsUJS.CLASS_NAME_ATTR + ']';
-      } else {
-        var selector = searchSelector + ' [' + window.ReactRailsUJS.CLASS_NAME_ATTR + ']';
+      var selector, parent;
+
+      switch (typeof searchSelector) {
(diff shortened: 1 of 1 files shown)


rmosolgo:
:+1: nice feature! 

Can you investigate that failing test? It looks like it passes on master. Does it fail locally?

Also, could you add a test for this new behavior? I think you could base your new test off of the existing one for mounting & unmounting by selector strings. 

You can see that the test clicks some buttons and makes sure that components are unmounted and re-mounted: https://github.com/reactjs/react-rails/blob/master/test/react/rails/view_helper_test.rb#L135-L136

The buttons are defined in this Rails view: https://github.com/reactjs/react-rails/blob/1c03b00b8105c78fbf674abe34be4941084bede4/test/dummy/app/views/pages/show.html.erb

So, I think you could make 2 new buttons which call those functions with DOM elements instead of strings. Then, in your test, click those buttons and make sure components are unmounted and re-mounted, just like the existing test.

ximus:
I fixed and added tests for this branch here: https://github.com/reactjs/react-rails/compare/master...ximus:support-dom-node-context

let me know what you want me to do with it

rmosolgo:
Awesome, please open a new Pull Request! If we merge that one, then we can close this one.

aratak:
Thanks, @ximus!

rmosolgo:
closed by #358 , Thanks for getting this started!

PR #341, opened 2015-08-25
Make server rendering more efficient

towry:
When server rendering, all components will be in components.js
and Execjs will evaluate the js code even some component is not
needed.

Add 'use_ssr' option, to only evaluate the component that specific by
controller. Note, when `use_ssr` is true, then there no need to provide
the `prerender: true` option.

```erb
react_component('Components.IndexView', {}, {use_ssr: true})
```

If `use_ssr` is not provided, then react-rails will works like before.

Here is an example: https://github.com/towry/rails-react-server-rendering-example.

### DIFF
--- a/lib/react/rails/engine.rb
+++ b/lib/react/rails/engine.rb
@@ -5,6 +5,11 @@ module React
         sprockets_env = app.assets || Sprockets # Sprockets 3.x expects this in a different place
         sprockets_env.register_engine(".jsx", React::JSX::Template)
       end
+
+      initializer 'react_rails.assets.precompile' do |app|
+        precompile = File.join("components", '*.ssr.js')
+        app.config.assets.precompile += [precompile]
+      end
     end
   end
 end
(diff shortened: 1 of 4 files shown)


rmosolgo:
One way that people have optimized this is make a `server_rendering.js` file which requires all components needed for server rendering. It's not included in `application.js`. It's only passed to `react-rails`, to be loaded into ExecJS. This way, they don't load _every_ component into ExecJS, only the subset which is needed for server rendering. 

Does this solution offer anything which that ^^ solution doesn't provide?

Another option for distributing this code would be to make  a custom renderer. You could make a Ruby gem which depends on `react-rails`, then in your gem, extend `SprocketsRenderer` with your desired behavior. 

`react-rails` accepts a config, `config.react.server_renderer`, where you could assign the new renderer class. That would cause `react-rails` to use your provided renderer instead of a built-in one.

towry:
In the second option, I have made a gem called `reactssr-rails` that it's just a renderer, but things goes complicated when I put this gem to our project (Maybe I have been working on the wrong way, but our project render all react components on server side).

I have a project that we render all react components on server side, so all the component goes into `server_rendering.js` file?  Because it seems that react-rails load everything in `files `:

```ruby
        files: ['react.js', 'server_rendering.js', 'components.js']
    }
```

PR #343, opened 2015-08-27
- Change version of installation to 1.2.0 in README so we dont run in…

vipulnsward:
…to installation issues with generator, due to sprockets


rmosolgo:
Could you elaborate a bit? I don't understand!

vipulnsward:
@rmosolgo If someone tries to run generator or using the gem on previous versions without the sprockets fix, it errors right now.

rmosolgo:
thanks!

PR #347, opened 2015-08-31
Reposition asset path build

### LINKED ISSUE #345: Put configuration in an initializer
I like to keep my configuration as separated into relevant files as possible, that's why I would love to be able to have an `/config/initializers/react.rb` file containing the `react-rails` configuration options.

However, due to the [`asset_paths` being added in the `config.before_initialize` hook of the railtie](https://github.com/reactjs/react-rails/blob/259e299286b4fa40c678ffb4dc64c9ddca9fde92/lib/react/rails/railtie.rb#L53) the `addon` option is ignored if it's set in an initializer instead of in `application.rb`.

Is it possible to move this asset path generation to the `after_initialize` hook, like the other stuff? I'm assuming it's not since you made a deliberate split of what goes into `before` and `after`, but I figured I'd make sure there is a reason before I start spending time maybe doing a PR to fix it :-)

dv:
The standard config/initializers are only run after `railtie#before_initialize`, but we have to build the asset paths before `railtie#after_initialize` since at that point Sprockets has already frozen the asset paths.

This PR moves the asset path build to a point in between `before_initialize` and `after_initialize`, after the standard initializers are run but before Sprockets freezes the environment.

Fixes #345

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -51,7 +51,7 @@ module React
 
       end
 
-      config.before_initialize do |app|
+      initializer "react_rails.set_variant", after: :engines_blank_point, group: :all do |app|
         asset_variant = React::Rails::AssetVariant.new({
           variant: app.config.react.variant,
           addons: app.config.react.addons,
--- a/test/dummy/config/application.rb
+++ b/test/dummy/config/application.rb
@@ -26,6 +26,7 @@ module Dummy
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
     config.react.variant = :production
+    config.react.addons = false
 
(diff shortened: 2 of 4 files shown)


dv:
I spent a few hours looking on how to best test the initializer code, but apparently this is almost impossible - since we would have to completely reboot the Rails app, and I couldn't find anywhere how to do that in a test.

The compromise I've used here is to hardcode an initializer in the dummy app, which overrides a setting set in the dummy app's `application.rb`. The single test that checks the path now has the added task of checking if the addons option was set before Sprockets froze the world.

rmosolgo:
Yeah, I remember trying to find some way to test it also, and basically giving up :P 

But the way you tested it looks good to me. And it passes on all our Ruby & Rails versions :D

rmosolgo:
Thanks! :confetti_ball:

PR #348, opened 2015-09-01
Add documentation for controller component rendering

garbles:
As per discussion in the [pull request for this feature](https://github.com/reactjs/react-rails/pull/329#issuecomment-136419566). :neckbeard: 

/cc @rmosolgo

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,22 @@ end
 - On MRI, you'll get a deadlock with `pool_size` > 1
 - If you're using JRuby, you can increase `pool_size` to have real multi-threaded rendering.
 
+### Rendering components instead of views
+
+Components can also be prerendered directly from a controller action with the custom `component` renderer. For example:
+
+```ruby
+class TodoController < ApplicationController
+  def index
+    @todos = Todo.all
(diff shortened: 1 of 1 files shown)


borisrorsvort:
Maybe we should say if there is a layout option or not :wink:

garbles:
@borisrorsvort thought about that, but I thought that stating `The component will be rendered in place of yield.` would be enough. As a renderer, it accepts all of the usual options for rendering a view including `layout`.

Could change this to `The component will be rendered in place of yield in the layout` if we think that it makes things slightly more clear.

borisrorsvort:
Or "it behaves the same as normal render and his usual options"

rmosolgo:
^^ I agree, signaling that "normal" render options still work would be a good addition

rmosolgo:
@garbles if/when you get a chance to update this, go ahead and remove the note about `~>1.3.0`. I'll cut a new version after merging, so I don't think it'll be necessary

garbles:
@rmosolgo @borisrorsvort thanks for the bump. I've pushed an update.

rmosolgo:
thanks!

PR #350, opened 2015-09-04
docs(readme) document some customization points

rmosolgo:
I hope this will help people understand & improve react-rails! And help them find ways to extend it in the ways they want to, eg Flux-on-the-server, gon.


vipulnsward:
I was playing around to use redux - https://github.com/vipulnsward/react-rails/commit/e240b6486c866296e6f91275256bf0aa60ca2f0f

I think, a new helper can be introduced if needed by someone, without affecting existing helper. 
More tricky is the ujs handling.

rmosolgo:
Oh, nice, that allows you to inject `react_redux_state`  into the page, then load it up? 

Yeah, I imagine if you need custom UJS you could just remove:

```js 
//= require react_ujs
```

And instead, require the custom UJS

```js 
//= require custom_ujs
```

PR #354, opened 2015-09-16
findDOMNodes(): Support jquery selectors

ximus:
In my use case, I need to initialize react components in a context referenced only by a jquery object. Trust me there is little I can do to get a CSS selector instead of this object in this use case I'm working on. However, `ReactRailsUJS.mountComponents()` Only supports string CSS selectors. 

There is no simple way to get a string CSS selector from a dom node or jquery object.

With much simplicity, this adds much value by enabling support for jquery objects passed to `ReactRailsUJS.mountComponents()`.

I hope you find this useful, feel free to re-work any way that suits you.

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -13,6 +13,10 @@
     // helper method for the mount and unmount methods to find the
     // `data-react-class` DOM elements
     findDOMNodes: function(searchSelector) {
+      if ($ && searchSelector instanceof $) {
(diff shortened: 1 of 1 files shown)


ximus:
another way of doing it, maybe more standard, could be to accept an array of DOM nodes and call `querySelectorAll` using `window.ReactRailsUJS.CLASS_NAME_ATTR` on each of them. Then flat map the results and use those as the nodes to initialize as react elements.

ximus:
I think https://github.com/reactjs/react-rails/pull/340/files does this more cleanly

rmosolgo:
Ok, that PR is stuck, feel free to either checkout his branch and fix the tests or re-implement yourself!

PR #358, opened 2015-09-27
Allow to pass DOM object as parent selector (second PR)

ximus:
Extend possibility of findDOMNodes and as a consequence mountComponents. It allows to use it with an element which was inserted to the DOM tree without react.js. Example:
```
$(document).on('nested:fieldAdded', function(e) {
  window.ReactRailsUJS.mountComponents(e.target);
});
```
Follow up from https://github.com/reactjs/react-rails/pull/340

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -26,7 +26,7 @@
           parent = searchSelector;
           break;
         case 'string':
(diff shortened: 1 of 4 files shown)


rmosolgo:
:+1: Thanks for getting this in shape! 

I guess it would misbehave if someone did this:

```javascript 
stringObject = new String("myString")
typeof stringObject // "object"
```

But, every JS guide I've ever seen just says "don't use `new String`", so I don't think we should worry about it!

PR #365, opened 2015-10-07
Update to React 0.14

rmosolgo:
This is kind of a non-pull request. I've used our existing update process, but it doesn't work :disappointed:  

The bower package doesn't include `ReactDOMServer`, which we need for server rendering.

Also, it's kind of a downer to introduce that second require:

```
//= require react
//= require react-dom
```

Because this is Ruby on Rails, I think it would be acceptable to assume that we want `react-dom`.

So, maybe we need a new update & build process:

- Get libraries from npm
- Build custom bundles for react-rails (including `react` & `react-dom`, `react-dom/server`?)

Any suggestions?


vipulnsward:
How are the ones listed on https://facebook.github.io/react/downloads.html generated?
We can just reuses the browser bundles from there?

rmosolgo:
Score! It looks like that includes `ReactDOMServer`, let me give it a try...

rmosolgo:
:weary: It includes `ReactDOMServer` so that it can attach deprecated functions to `React`, but it doesn't export the object

vipulnsward:
I think we can ship with react-rails.js and that should henceforth include whatever is needed to get react running. 

```
// react-rails.js

//= require react
//= require react-dom
```

And in Application - 
```
//= require react-rails
```

IvRRimum:
When 0.14 React will be available in this gem ? I developoing rails server, but i dont want to use outdated react version.

onpaws:
@rmosolgo Thanks for your effort on getting React 0.14 in place. While I'm (relatively) new to React and this gem if you need help getting this sorted out I'd be happy to dive in.

rmosolgo:
@onpaws I'm hoping to take a pass at this on Sunday or Monday, but if you're interested, definitely take a shot at it!

Here's what I've got in mind:

### Goals

- Include `react` & `react-dom` on the client
- Include `react`, `react-dom` and `react-dom/server` on the server
- Preserve the simple `//= require react` API inside sprockets JS files (or perhaps `react-rails` as @vipulnsward  described above)
- Preserve the ability to drop in a local copy of React if you want to override the `react-rails` default
- Preserve the ability to update the React version in the gem source by running a Rake task (replacing the current `rake react:update`)

### Method 

I'm open to anything that accomplishes the goals above. I haven't tried it yet, but this is what I'd try:

- Add a new directory to this project which is a webpack project
- Make a couple of manifest files for `react-rails` (one for the client, one for the server) using Node.js `require` API to get the right packages 
- Set up the webpack config to build those manifests into the existing `lib/assets/react-source` directories (setting `NODE_ENV=production` for the production build)
- Make a rake task that runs webpack for that build 

### Not sure about 

- How to include add-ons? I see they're all in their own modules now. I'd like to bundle those in the `-with-addons` files.
- At some point, I'd like to make it possible to update the React.js version without updating the gem. Maybe supporting React 0.14 will shed some light on how to add that feature.

onpaws:
Awesome! Your plan sounds great. 
Resolving an 'invariant violation' error on my app today took up quite some time today  - hoping to continue on this tomorrow/Monday.

Regarding your last item, I noticed angularjs-rails appears to offer something similar to what you describe. Perhaps this might be helpful?
https://github.com/hiravgandhi/angularjs-rails/blob/master/tasks/angularjs-rails/stable_updater.rb

rmosolgo:
I took a pass at this here: #376

PR #366, opened 2015-10-09
Support ReactDOM in react_ujs

axsuul:
In 0.14, we now have `ReactDOM`. Prefer it for now to get rid of errors in the console.


rmosolgo:
:+1: great, at least we can support drop-in React 0.14 for the time being. Thank you!

Issue #353, opened 2015-09-15
React Rails Testing


taylorallred:
I just updated to 1.3.0, ran `rake test` and got this error message: 

`ActionView::Template::Error: undefined method 'react_component' for nil:NilClass
    app/views/layouts/application.html.erb:30:in '_app_views_layouts_application_html_erb___3811361224884677630_70195218569900'`

This only happens when I run my tests and doesn't happen when I run the code in the browser. When I stepped back to 1.2.0 it fixed the issue.

I'm calling `react_component` in my application.html.erb file.

rmosolgo:
sounds like I broke something, hmm

rmosolgo:
Ah, is this only in test. I think my implementation depends on Rack middleware doing its thing, let me find a fix for this ...

rmosolgo:
could you share the test that's breaking? The tests in this repo pass but I guess there's another kind of test that I missed!

taylorallred:
The most basic test that is failing is:
`  test '.show should succeed' do
    get :show
    assert_response :success
  end`

rmosolgo:
what kind of test is that? what test framework? is it inside a class which extends a base test class? is it part of a `describe` block?

rmosolgo:
(seems like this middleware is not applied in the test context: https://github.com/reactjs/react-rails/blob/master/lib/react/rails/render_middleware.rb)

taylorallred:
We use:
group :development, :test do
  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
  gem 'byebug'

  # Shim to load environment variables from .env into ENV in development. Read more: https://github.com/bkeepers/dotenv
  gem 'dotenv-rails'

  # Letter opener: https://github.com/ryanb/letter_opener
  gem 'letter_opener', :group => :development

  # Headless phantomjs(chrome based browser) for javascrip testing
  gem 'poltergeist', :require => false

  # Ruby Pry, debug ruby https://github.com/deivid-rodriguez/pry-byebug
  gem 'pry-byebug'

  # Static syle checking https://github.com/bbatsov/rubocop
  # version 0.33.0 is causing issues -> when running autocomplete it find multiple infinite loops
  # Untill we solve that problem, we'll go back to the last stable version.
  gem 'rubocop', '0.32.1', :require => false

  # Optional code coverage
  gem 'simplecov', :require => false

  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
  gem 'spring'

[...]

PR #367, opened 2015-10-09
fix for rspec

masarakki:
in view tests of rspec, we must assign vars by myself.


lib/react/rails/railtie.rb
```diff
@@ -83,6 +83,10 @@ class Railtie < ::Rails::Railtie
         React::ServerRendering.reset_pool
         # Reload renderers in dev when files change
         ActionDispatch::Reloader.to_prepare { React::ServerRendering.reset_pool }
+        if defined? RSpec::Rails
+          require 'react/rails/test_helper'
```
rmosolgo: Would it hurt to _always_ require this file? It's not Rspec-specific, right? So we could move it to `react/rails.rb` along with the other `require` calls. 

Also, if this helper is specific to Rspec, what do you think about calling it `RspecHelper` instead of `TestHelper`? I think that expresses its purpose a little bit more.

rmosolgo: Or, maybe `ViewTestHelper` if this also works with Minitest

rmosolgo:
Thanks for adding this fix! Could you also add a test for this behavior? 

I see that you were using Rspec, but this project uses Minitest. Can you test this behavior by adding an `ActionView::TestCase` to this project? That would be good because it would make sure that we don't break this in the future!

masarakki:
I have no idea to test it without rspec...

rmosolgo:
Yes, that would be fine!

masarakki:
add rspec tests

rmosolgo:
Hmm, not sure what's wrong with this build. 

It turns out this problem is part of a bigger issue. I was able to reproduce the issue using `ActionView::TestCase` and I've suggested a fix here: https://github.com/reactjs/react-rails/pull/375

Do you want to try that branch? If it works for you, I'll merge it, since I think it also fixes the issue described here: https://github.com/reactjs/react-rails/issues/353#issuecomment-148847051

masarakki:
I think it's just a travis's problem...
rebuilding with dummy commit

rmosolgo:
Sorry @masarakki , I think this problem was larger than just Rspec tests.  It was also broken for:

- `SimpleForm::Inputs`: https://github.com/reactjs/react-rails/issues/377
- `cells` gem: https://github.com/reactjs/react-rails/issues/353#issuecomment-148847051

So, I refactored the helper a bit and used `ActionView::TestCase` to ensure that the `react_component` helper could be used these different ways: https://github.com/reactjs/react-rails/pull/375

I prefer that option because it's less code overall and it handles more of those broken cases. Could you check that it works for you, too? I've just released it in version `1.3.3`.

masarakki:
v1.3.3 works well. great job!!

PR #376, opened 2015-10-18
feat(JS) add webpack build process

rmosolgo:
Here's my best guess at how to update to React 0.14:

- Write manifests for server/browser, no-addons/with-addons 
- Build the manifests with Webpack 
- Copy them into the gem source, just like the old React.js files 
- Longer discussion: https://github.com/reactjs/react-rails/pull/365

Other changes:

- Server rendering uses `ReactDOMServer` and `react-server.js`
- Split the rake task into two (`update` = `build` + `copy`)

Lol: 
image


rmosolgo:
@zpao Could you take a look at the JavaScript side of this? I think this is an OK way to use the new JS modules in the browser & server, but I'm a bit out of my element!

rmosolgo:
looks good in my own app, here goes!

zpao:
Sorry for the delay, this looks good to me!

PR #378, opened 2015-10-22
react-addons-linked-state-mixin

SylarRuby:
Could you include this add on in your next update? It would be really useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,6 @@
 
 # react-rails
 
-> ## react-rails version disclaimer
-
-> ***This README is for `1.x` branch which is still in development. Please switch to latest `0.x` branch for stable version.***
-
-> ***Additionally: `0.x` branch directly follows React versions, `1.x` will not do so.***
 
 `react-rails` makes it easy to use [React](http://facebook.github.io/react/) and [JSX](http://facebook.github.io/react/docs/jsx-in-depth.html) in your Ruby on Rails (3.1+) application. `react-rails` can:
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
I see your code includes a the tilt dependency, but the issue title is the name of a React addon. Is one of those things a typo?

I'm happy to merge the tilt dependency, I wonder how it got left behind. Just making sure this is the right thing!

SylarRuby:
@rmosolgo Hi. I'm new to pull requests so sorry for asking this question here. I want `react-addons-linked-state-mixin`. Is that possible?

rmosolgo:
No problem, thanks for clarifying. Usually, you can report a bug or request a new feature by opening an _issue_ instead of a _pull request_. 

I expect `1.4.0` (just released today) to contain `React.addons.LinkedStateMixin`: https://github.com/reactjs/react-rails/blob/master/react-builds/addons-object.js#L4

You can activate mixins in your rails app by adding a config:

```ruby 
# in application.rb
  # ... after other configs ... 
  config.react.addons = true 
end
```

Then, restart your development server. `React.addons` should be defined in the browser. 

Does that work for you?

SylarRuby:
I am using Rails 4 and in my application.rb I have:
```
module App
  class Application < Rails::Application
  [...]
  config.react.addons = true
  end
end
```

Is that correct @rmosolgo ?

rmosolgo:
Yep, that looks right!

SylarRuby:
Well then it works great. This can be closed now. Thanks

rmosolgo:
Glad to hear it!

PR #383, opened 2015-10-27
Code taken from patch done by ownerscircle fork

rstudner:
https://github.com/reactjs/react-rails/compare/master...ownerscircle:perf_testing

I tested this pretty thoroughly locally, but I'm also not a rails expert so i'd love for someone else to pull this down and try it in a production/pre-compiled environment.

### DIFF
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -11,11 +11,20 @@ module React
         js_code = CONSOLE_POLYFILL.dup
 
         filenames.each do |filename|
-          js_code << ::Rails.application.assets[filename].to_s
+          js_code << load_asset(filename)
         end
 
         super(options.merge(code: js_code))
       end
+
+      def load_asset(file)
+        if ::Rails.application.config.assets.compile
+          ::Rails.application.assets[file].to_s
+        else
+          asset_path = ActionView::Base.new.asset_path(file)
+          File.read(File.join(::Rails.public_path, asset_path))
+        end
+      end      
 
       def render(component_name, props, prerender_options)
         # pass prerender: :static to use renderToStaticMarkup


rmosolgo:
Awesome! It would be great to get a test for this behavior, too. I think you could: 

- Add an already-compile asset file to the dummy app in `test/dummy/public/{??}`
- Add a test to sprockets_renderer_test.rb: 

  ```ruby
  def test_uses_compiled_asset_files 
    Rails.application.config.assets.compile = false 
    precompiled_renderer = SprocketsRenderer.new(files: ["..."])
    result = precompiled_renderer.render("SomeComponent", {}) # based on contents of asset
    assert_includes(result, "something") # whatever should be there 
  ensure 
    Rails.application.config.assets.compile = true
  end
  ```
 
- One possible refactor: instead of reaching for `Rails.application.config.assets.compile` inside `#load_asset`, you could accept an option to `#initialize` and store it in an instance variable. The option would default to `Rails.application.config.assets.compile`, but you could override it for testing without having to mess with the configs.

rstudner:
I freely admit I was both unable to get all the tests to even run as well as not being able to get any of the pre-existing components TodoList etc to render/pre-render a single time so I could use one of them as the public/<test component>

So not sure if I can proceed (but I of course 100% agree with the adding of the test etc hah)

(i'm using rubymine, and 100% of all other rails/gem projects i've ever imported this all just worked fine hah)

rstudner:
Would this be an accurate test?
```
  def test_uses_compiled_asset_files
    Rails.application.config.assets.compile = true
    precompiled_renderer =  React::ServerRendering::SprocketsRenderer.new(files: ["react-server.js", "components/Todo.js"])
    result = precompiled_renderer.render("Todo", {todo: "get a real job"}, {}) # based on contents of asset
    assert_includes(result, "get a real job") # whatever should be there
  end
```

To note, if I have the initial compile = false it all fails.  So I changed that to 'true' and removed the ensure line.

rstudner:
@rmosolgo I'm writing up a test but having an issue digestion your suggested test:

```
precompiled_renderer = SprocketsRenderer.new(files: ["..."])
```

What do I put here?  if I try to put react.js or anything, it says it can't find the file (and that kills the test)

rmosolgo:
Oh, i guess you could just leave it blank and use the default `files` option: https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/sprockets_renderer.rb#L10

Yes, those files aren't there yet. Normally, they'd be moved to that location by `rake assets:precompile`. You could run that on the dummy app, for example:

```
$ cd test/dummy 
$ rake assets:precompile
```

Then it might find them!

rstudner:
@rmosolgo  A few points:

1. I had changed it to try the default files option already -- seemed like the right step.
2. I didn't think it was legit? for me to require a manual rake assets:precompile before running the tests -- but that makes sense of course. (but then wouldn't rake test have to be overridden to do this for this test? for a travis etc environment?)
3. the assets get put in react-rails/test/dummy/tmp/cache/assets/test/sprockets/v3.0/
4. the code still is trying to just look in react-rails/test/public/assets
5. the filenames of course are like: zOx_HJxZ2xP7SL8yNvetrX_Dn41YDxQYSLjOkxwlvD0.cache 

I guess this leads me to believe the patch code from earlier etc doesn't work at all and this is a minor fools errand :)

Thoughts?

rmosolgo:
I don't think it's a fool's errand, but it's just unfamiliar territory for me! I'm sure it's possible to do this "right", but it involves digging around in a bunch of Sprockets APIs to figure out how to make it work :grimacing:

rstudner:
I'm going to illicit some help from my Rails expert cohorts (i'm the scala, clojure, ember, angular, backbone, react person) haha.

Thanks for all the help/feedback, hopefully I can get this thing working soon :)

rstudner:
@rmosolgo btw - closed this for 2 reasons.  1) was just very stale/already pushed/I love rebase/didn't want to force push haha  2) Found another approach with the help of a coworker/friend (who i'll get his public github and get him the credit asap).. anyway -- going to push up a new branch/PR with this done another way w/ a test etc.

dfguo:
I'm having the same issue where my assets are not found in ::Rails.application.assets[filename].to_s. This change actually fixes it.

rstudner:
@dfguo hrm.  How are you using/testing it?  The issue for me, is that once I "go to prod" my application-hash.js "has react in it" (and everything else right :)?  and my server side rendering is trying to "find react.js" and I haven't been able to figure out (using the manifest cache/anything) how to resolve that.

do you have anything like this in your applications.rb:

```
    #Make sure server side react components get at least this stuff to use (moment & i18n being key)
        files: ["react.js", "react-server.js", "i18n.js", "momentjs", "components.js", "portal.js"], # files to load for prerendering
        replay_console: true,                 # if true, console.* will be replayed client-side
    }
```

dfguo:
@rstudner yes, similar problem. I'm not sure why in prod my file returns as nil when calling ``` ::Rails.application.assets[filename].to_s```.  I took your code and added a simple cache:

```
      def load_asset(file)
        if ::Rails.application.config.assets.compile
          ::Rails.application.assets[file].to_s
        else
          @filemap[file] ||= read_file(file)
        end
      end

      def read_file(file)
        asset_path = ActionView::Base.new.asset_path(file)
        File.read(File.join(::Rails.public_path, asset_path))
      end
```

Take note that my set up is slightly different, I use webpack to generate one final bundle file:

```
Rails.application.config.react.server_renderer_options = {
    files: ['webpack-bundle.js'], # files to load for prerendering
}
```

rstudner:
Ahh yeah, I'm almost to using webpack with this project and then i'd do the exact same as above.  So when you *arent* using webpack, the problem is that it looks for the files in that list, and of course, they don't exist (since they've been Sprocketed hah).

Have you seen the gem react_on_rails?  Solid support for doing webpack w/ rails etc.  I didn't realize you could easily do webpack w/ rails and still use this particular gem (i'm behind in reading the docs, clearly)

How do you have webpack configured to work with your rails app and this gem (if you don't mind sharing)

dfguo:
@rstudner it's actually fairly simple. We did a sharing at Ruby Conf China this year. 

Video: https://www.youtube.com/watch?v=3G4SFsl5p6s
Slides: https://speakerdeck.com/onlimii/rails-off-assets-pipeline

TLDR version: output the webpack generated bundle to app/assets/javascript/bundle  folder and let Rails read it. 

We have been using react-rails gem purely for server-side rendering. Honestly, it's a bit bloated for what we use it for.

PR #384, opened 2015-10-27
Add ability to render custom attributes from the controller

kaiwood:
Currently, all custom attributes are sliced from options  when using the controller renderer, instead of forwarding it to the view helper, making it impossible to set a custom class name or id.

This patch changes the behaviour to act more like render_component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,13 +212,14 @@ Components can also be prerendered directly from a controller action with the cu
 class TodoController < ApplicationController
   def index
     @todos = Todo.all
-    render component: 'TodoList', props: { todos: @todos }, tag: 'span'
+    render component: 'TodoList', props: { todos: @todos }, tag: 'span', class: 'todo'
   end
 end
 ```
 
 This custom renderer behaves the same as a normal view renderer and accepts the usual arguments - `content_type`, `layout`, `location` and `status`.
-By default, your current layout will be used and the component, rather than a view, will be rendered in place of `yield`.
+By default, your current layout will be used and the component, rather than a view, will be rendered in place of `yield`. Custom data-* attributes
+can be passed like `data: {remote: true}`.
 
 ### Component generator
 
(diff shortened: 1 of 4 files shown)


rmosolgo:
Very nice, I definitely agree we should allow more options to be passed through the controller renderer! 

But right now it's pushing through a bunch of `ActionController::Base#render` options, too: 

```
<span 
  class=\"custom-class\" 
  data-react-class=\"TodoList\" 
  data-react-props=\"{&quot;todos&quot;:[&quot;Render this inline&quot;]}\" 
  id=\"custom-id\" 
  layout=\"#&lt;Proc:0x000000038db4b0@/home/travis/.rvm/gems/ruby-2.2.0/gems/actionpack-3.2.22/lib/abstract_controller/layouts.rb:383&gt;\" 
  prefixes=\"server application\" 
  props=\"{:todos=&gt;[&quot;Render this inline&quot;]}\" 
  template=\"inline_component\"
>
```

(I grab this from the Travis CI run)

What do you think about just adding some more options to the whitelist?

kaiwood:
Extending the whitelist was my first thought, as my own intention was to have access to the class and id attribute for styling. 

But shouldn't at least data-* or aria-* attributes whitelisted too? Might not be to far fetched that someone needs access to those, e.g. for some kind of plugin or accessibility thingy...

I'll take a look at it, didn't mean to break Travis anyway. :)

rmosolgo:
That's a good point, I wonder if just whitelisting `:data, :aria` would do the trick, since you can often pass those like `data: { remote: true }`

kaiwood:
Ok, i restored the whitelisting and added class, id, data and aria.

Stuff like `data: {remote: true}` works as expected. The thing with aria-* is, that the nested syntax was added in https://github.com/rails/rails/commit/a794fd2c0a99e7e917f6dbf83d781f621a39f3ad and is only supported in Rails 4.2 upwards…

We would need to add all 35 aria-attributes from the spec to the whitelist for the versions prior, looks like a little bit over the top to me. I think we can get away with this in the current form :)

rmosolgo:
:tada:  Looks great! Down the road, if someone needs more attributes than this, we could extract a `REACT_COMPONENT_OPTIONS` array which people could mutate as they see fit. 

Thanks for this improvement!

PR #385, opened 2015-10-28
Switch to the new Sprockets processor

eugenijusr:
Starting version 4 Sprockets is removing the `register_engine` method which registers JSX transformers using Tilt based templates breaking this gem for the current Rails 5 `master`. This has been deprecated since Sprockets version 3:
https://github.com/rails/sprockets/blob/master/UPGRADING.md

This PR adds a JSX Processor and registers it via `register_processor` as `application/jsx` mime type. Is there a reserved mime type for JSX templates?

### DIFF
--- a/lib/react/jsx/processor.rb
+++ b/lib/react/jsx/processor.rb
@@ -1,16 +1,8 @@
 module React
   module JSX
     class Processor
-      def self.instance
-        @instance ||= new
-      end
-      
       def self.call(input)
-        instance.call(input)
-      end
-
-      def call(input)
-        {data: JSX::transform(input[:data])}
+        JSX::transform(input[:data])
       end
     end
   end
--- a/lib/react/jsx/processor.rb
+++ b/lib/react/jsx/processor.rb
@@ -1,0 +1,17 @@
+module React
+  module JSX
+    class Processor
+      def self.instance
+        @instance ||= new
+      end
+      
+      def self.call(input)
+        instance.call(input)
+      end
+
+      def call(input)
+        {data: JSX::transform(input[:data])}
+      end
+    end
+  end
+end
--- a/lib/react/rails/engine.rb
+++ b/lib/react/rails/engine.rb
@@ -3,7 +3,12 @@ module React
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
         sprockets_env = app.assets || Sprockets # Sprockets 3.x expects this in a different place
-        sprockets_env.register_engine(".jsx", React::JSX::Template)
+        if Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new("3.0.0")
+          sprockets_env.register_mime_type("application/jsx", extensions: [".jsx", ".js.jsx"])
+          sprockets_env.register_transformer("application/jsx", "application/javascript", React::JSX::Processor)
+        else
+          sprockets_env.register_engine(".jsx", React::JSX::Template)
+        end
       end
     end
   end
--- a/lib/react/jsx.rb
+++ b/lib/react/jsx.rb
@@ -1,4 +1,5 @@
 require 'execjs'
+require 'react/jsx/processor'
 require 'react/jsx/template'
 require 'react/jsx/jsx_transformer'
 require 'react/jsx/babel_transformer'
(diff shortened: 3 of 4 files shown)


lib/react/jsx/processor.rb
```diff
@@ -0,0 +1,17 @@
...
+      def self.instance
+        @instance ||= new
+      end
+      
+      def self.call(input)
+        instance.call(input)
+      end
+
+      def call(input)
+        @output ||= JSX::transform(input[:data])
```
rmosolgo: Hmmm as you can see in the specs, this `@output` is getting assigned in the first use, but then never being reassigned. This causes all inputs to have the same output! 

Is there a downside to simply 
```
class Processor 
  def self.call(input)
    JSX::transform(input[:data])
  end 
end
```

?

rmosolgo:
Thanks for updating us! 

I don't think there is a  special mime type for `.jsx`. They're usually requested & served as JavaScript files, so I think the normal JS mime type is good enough!

lib/react/jsx/processor.rb
eugenijusr: I think this was done for performance reasons in the Tilt template version of this. For these Sprocket processors there is a way to do a similar thing by implementing an optional `cache_key` method.

lib/react/jsx/processor.rb
```diff
@@ -0,0 +1,17 @@
...
+      def self.instance
+        @instance ||= new
+      end
+      
+      def self.call(input)
+        instance.call(input)
+      end
+
+      def call(input)
+        {data: JSX::transform(input[:data])}
```
rmosolgo: from https://github.com/rails/sprockets#return-hash:

> Returning a String is shorthand for returning `{ data: str }`

Can we return a string only? Do we need any metadata?

lib/react/jsx/processor.rb
```diff
@@ -0,0 +1,17 @@
+module React
+  module JSX
+    class Processor
+      def self.instance
+        @instance ||= new
+      end
+      
+      def self.call(input)
+        instance.call(input)
```
rmosolgo: Do we benefit from having `@instance` ? What if this class method was simply 

```ruby
def self.call(input)
  JSX::transform(input[:data])
end
```

would that be sufficient?

lib/react/rails/engine.rb
```diff
@@ -3,7 +3,12 @@ module Rails
     class Engine < ::Rails::Engine
       initializer "react_rails.setup_engine", :group => :all do |app|
         sprockets_env = app.assets || Sprockets # Sprockets 3.x expects this in a different place
-        sprockets_env.register_engine(".jsx", React::JSX::Template)
+        if Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new("3.0.0")
+          sprockets_env.register_mime_type("application/jsx", extensions: [".jsx", ".js.jsx"])
```
rmosolgo: Is this for handling requests with mime-type `application/jsx` ?  Where do those come from? 

For react-rails, I think all requests are `application/javascript`. If that's correct, let's remove this line since we don't need it!

eugenijusr: This is only the internal Sprockets mime-type definition. That's the way Sprockets maps extensions with transformers - via mime-types.

```ruby
register_mime_type arbitrary_mime_type, extensions: list_of_extensions
register_transformer arbitrary_mime_type, target_mime_type, SomeProcessor
```

https://github.com/rails/sprockets/blob/master/lib/sprockets.rb

Starting version 4 you will no longer be able to map extensions directly to transformers so the mime-type is actually required. We cannot use the existing `application/javascript` because that's the target mime-type reserved for vanilla JS files.

Their 2->3 upgrade doc even says:
*Register a content type for file extension, its okay if its made up*

Alternatives to `application/jsx` would be to:

- Follow the ERB convention and map JSX files to `application/javascript+jsx` mime-type similar to the way Sprockets maps `.html.erb` files to `text/html+ruby`.
- Follow the CoffeeScript convention and map JSX files to `text/jsx` mime-type similar to the way Sprockets maps `.js.coffee` to `text/coffeescript`.

lib/react/jsx/processor.rb
eugenijusr: True, there's no benefit if there's no caching.

lib/react/rails/engine.rb
rmosolgo: Thanks for the explanation! I see why you chose to do it this way and it sounds good to me!

rmosolgo:
:sparkles: Thanks again for getting us up-to-date (and getting _me_ up to date :stuck_out_tongue_closed_eyes:) !

PR #388, opened 2015-10-29
Update README to hint for newer syntax

kaiwood:
Since React 0.12 the ES6 style syntax can be used with CoffeeScript classes as well. The docs should give a hint for this.


rmosolgo:
Fair enough, I'm not into Class-style components but, hey, you might as well inform people :P

Thanks, you're on a roll! :)

PR #389, opened 2015-10-29
Fix ExecJSRenderer interface

hokaccha:
I'd like to use `ExecJSRenderer` for server side rendering as below.

```ruby
config.react.server_renderer = ExecJSRenderer
config.react.server_renderer_options[:code] = File.read "path/to/compiled_component.js"
```

However, it dose not work because

* `ExecJSRenderer#render`'s `prerender_options` expects a Hash , but actually it's `true` or `:static`.
* `ExecJSRenderer#render` dose not convert to JSON when `props` is not string.

This patch will resolve it.

### DIFF
--- a/lib/react/server_rendering/exec_js_renderer.rb
+++ b/lib/react/server_rendering/exec_js_renderer.rb
@@ -11,7 +11,17 @@ module React
       end
 
       def render(component_name, props, prerender_options)
-        render_function = prerender_options.fetch(:render_function, "renderToString")
+        # pass prerender: :static to use renderToStaticMarkup
+        render_function = if prerender_options == :static
+            "renderToStaticMarkup"
+          else
+            "renderToString"
+          end
+
+        if !props.is_a?(String)
+          props = props.to_json
+        end
+
         js_code = <<-JS
           (function () {
             #{before_render(component_name, props, prerender_options)}
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -17,21 +17,6 @@ module React
         super(options.merge(code: js_code))
       end
 
-      def render(component_name, props, prerender_options)
-        # pass prerender: :static to use renderToStaticMarkup
-        react_render_method = if prerender_options == :static
-            "renderToStaticMarkup"
-          else
-            "renderToString"
-          end
-
-        if !props.is_a?(String)
-          props = props.to_json
-        end
-
-        super(component_name, props, {render_function: react_render_method})
-      end
-
       def after_render(component_name, props, prerender_options)
         @replay_console ? CONSOLE_REPLAY : ""
       end
--- a/test/dummy/app/controllers/server_controller.rb
+++ b/test/dummy/app/controllers/server_controller.rb
@@ -1,4 +1,16 @@
 class ServerController < ApplicationController
+  DUMMY_IMPLEMENTATION = "
+  var Todo = null
+  var React = {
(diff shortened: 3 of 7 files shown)


rmosolgo:
I don't think of `ExecJSRenderer#render` as _broken_, I think of it as _different_ (simpler, maybe even better)!  I like it because the input types are consistent:

- `props` is _always_ a string of JSON  (not a string-or-something-that-responds-to-to_json)
- `prerender_options` is _always_ a hash (not a boolean-or-symbol)

This means you have to call it differently from the view: 

```erb
<%= react_component("Todos", {todos: ["Thing 1", "Thing 2"]}.to_json, prerender: {render_function: "renderToStaticMarkup"}) %>
```

Of course, this is quite a mouthful, so I can think of a couple of ways to reduce that boilerplate code. Would either of these work for you?

### Custom helper

You could define a view helper that transforms inputs in the same way `SprocketsRenderer` does 

```ruby
# Prepare the arguments, then use the renderer 
def execjs_react_component(name, props, prerender=nil)
  # ensure that the props are strings 
  props_string = props.is_a?(String) ? props : props.to_json
  # prepare the prerender options 
  if prerender.present?
    prerender_function = prerender == :static ? "renderToStaticMarkup" : "renderToString"
    prerender_options = {render_function: prerender_function}
  else 
    prerender_options = false 
  end
  react_component(name, props_string, prerender: prerender_options)
end
```

Then you could use it like the other renderer:

```erb
<%= execjs_react_component("Todos", {todos: ["Thing 1", "Thing 2"]}, prerender: :static) %>
```

### Custom renderer 

Alternatively, you could extend `ExecJSRenderer` and override `#render` to perform the same input manipulation that `SprocketsRenderer` does.

### Also ...

You could probably make this a bit simpler by moving this operation to the beginning `react_component` method: 

https://github.com/reactjs/react-rails/blob/master/lib/react/rails/component_mount.rb#L35

Then you wouldn't have to make the `is_a?(String)` check yourself. That would be a nice patch for react-rails, I think!

(The only downside of this is that if someone made a render which _didn't_ want a string, they'd have to reimplement the helper. But as far as I know, there are no other custom renderers out there!)

### Goals 

Here are my goals in the server renderers: 

- Keep `ExecJSRenderer` as simple as possible (just coordinating ExecJS + React.js)
- Push as much Rails-specific logic into `SprocketsRenderer`
- Maintain a flexible interface for extending existing renderers or writing new ones (this is why I have `prerender:` as a Hash, it is more extensible)
- Provide a simple, pretty API for the most common uses 


What do you think? Will any of those other solutions work in your case? Or have I missed something?

hokaccha:
@rmosolgo Thank you for you reply!

> * Keep ExecJSRenderer as simple as possible (just coordinating ExecJS + React.js)
> * Push as much Rails-specific logic into SprocketsRenderer
> * Maintain a flexible interface for extending existing renderers or writing new ones (this is why I have prerender: as a Hash, it is more extensible)
> * Provide a simple, pretty API for the most common uses

I agree. However I believe that interface of `react_component` should not be changed by `Renderer` option.

This behavior is confusing:

```erb
SprocketsRenderer:
<%= react_component("Todos", {todos: ["Thing 1", "Thing 2"]}, prerender: true) %>

ExecJSRenderer:
<%= react_component("Todos", {todos: ["Thing 1", "Thing 2"]}.to_json, prerender: {render_function: "renderToStaticMarkup"}) %>
```

What do you think?

hokaccha:
I think another way. thanks.

PR #409, opened 2015-11-22
Snake-Case key to Camel-Case key for props when server render

ZhangHanDong:
Camle-Case in es6.jsx:

```
static propTypes = {
		form: PropTypes.object,
		errorMessages: PropTypes.array,
		defaultFieldValues: PropTypes.object
	}
```

I must be camle-case  in Rails: 

```
props = {form: {action: "new"}, errorMessages: ["errors1, errors2"], defaultFieldValues:  {some: "value"}}
```

I think it's better in Rails:

```
props = {form: {action: "new"}, error_messages: ["errors1, errors2"], default_field_values: {some: "value"}}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,14 @@ end
 - On MRI, you'll get a deadlock with `pool_size` > 1
 - If you're using JRuby, you can increase `pool_size` to have real multi-threaded rendering.
 
+You can configure camelize_props option and pass props with an underscored hash from rails but get a camelized hash in jsx :
+
+```ruby
+MyApp::Application.configure do
+  config.react.camelize_props = true #default false
+end
+```
+
 ### Rendering components instead of views
 
 Components can also be prerendered directly from a controller action with the custom `component` renderer. For example:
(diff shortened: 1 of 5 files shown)


rmosolgo:
Hi! Thanks for adding this feature. I can see how useful this will be!

How about making this an opt-in feature? I can think of two ways to make this opt-in:
- Add a config, like `config.camelize_props = true` 
- Make the view helper public & add documentation so that people can use it if they want.

I don't want to make it the _default_ behavior for two reasons:
- It will be backwards-incompatible for current users 
- I can imagine it would be unpredictable for some users (it's strange to pass in an underscored hash but get a camelized hash). 

Thanks again, how does that sound?

ZhangHanDong:
sounds good！ I'll be modify the code.

ZhangHanDong:
@rmosolgo 

I have added a config option:

```ruby

  config.react.camelize_props = true #default false
end

```

please check, thanks!

rmosolgo:
:tada: thanks, this is great!

ZhangHanDong:
@rmosolgo  thanks

yoonsy:
:+1:

Issue #395, opened 2015-11-06
Pre-compilation for es6 breaks with version 1.4.2


dkln:
Upgraded to react-rails gem version 1.4.2 this morning. This unfortunately breaks pre-compilation. es6/jsx files aren't picked up anymore (1.4.1 works fine btw).

 I suspect this is due changes made in the Rails Engine?

rmosolgo:
Maybe so. How do you set up es6 compilation? If you show me how you do that, I can add a test & try to get it going again!

dkln:
Well :smile:, This is kinda how my Gemfile looks like:

* react-rails
* sprockets
* sprockets-es6

So my `components.js` (that lives in `app/assets/javascripts`) contains something like this:

`components.js`:

```
//= require ./video
```

The filename of video, is in this case: `video.es6.jsx`. That used to work but now it can't find a suitable format anymore. To be honest, I did not try to change the extension (`.jsx.es6` or `.jsx.js` etc.).

rmosolgo:
I think this is the breaking change: https://github.com/reactjs/react-rails/pull/385 But I'm not sure exactly _how_ it broke things. 

In your case, do you get an error message? If so, what is the message and what is the stack trace? 

Or does it simply fail to transform?

dkln:
Yes I get an error message:

```
Sprockets::FileNotFound: couldn't find file './video' under 'website/app/assets/javascripts' with type 'application/javascript'
```

And the infamous stacktrace:

```
/Users/the-user/Projects/website/app/assets/javascripts/components.js:4
/Users/the-user/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/sprockets-3.4.0/lib/sprockets/resolve.rb:64:in `resolve!'
/Users/the-user/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/sprockets-3.4.0/lib/sprockets/directive_processor.rb:399:in `resolve'
/Users/the-user/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/sprockets-3.4.0/lib/sprockets/directive_processor.rb:207:in `process_require_directive'
/Users/the-user/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/sprockets-3.4.0/lib/sprockets/directive_processor.rb:180:in `block in process_directives'
[...]

PR #411, opened 2015-11-23
Fix ECMAScript compilation

eugenijusr:
Fixes https://github.com/reactjs/react-rails/issues/395 by adding `.es.jsx` and `.es6.jsx` extensions.

See https://github.com/reactjs/react-rails/issues/395#issuecomment-159058805 for details.


rmosolgo:
:+1: thanks for following up on this!

PR #417, opened 2015-12-03
Include CHANGELOG in gem

dbackeus:
Help developers know what's new using `bundle open react-rails` etc after a `bundle update` instead of having to navigate to github.


rmosolgo:
Happy to merge this, but  I think you need to sign the CLA first!

dbackeus:
@rmosolgo thanks for the heads up, I've clicked the "agree" button on that page now so I hope everything's ok?

rmosolgo:
Yep, thanks!

PR #427, opened 2015-12-21
Only require rails as a development dependency

natemontgomery:
Requiring Rails as a runtime dependency causes issues if application does not use ActiveRecord.

### DIFF
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'coffee-script-source', '~>1.8'
   s.add_dependency 'connection_pool'
   s.add_dependency 'execjs'
-  s.add_dependency 'rails', '>= 3.2'
+  s.add_dependency 'railties', '>= 3.2'
   s.add_dependency 'tilt'
   s.add_dependency 'babel-transpiler', '>=0.7.0'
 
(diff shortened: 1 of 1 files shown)


danchapman:
:+1:

vipulnsward:
Doesn't this have issues with sprockets not being available?

rmosolgo:
I think our usage of sprockets can be avoided if you're not on rails. Most of it is through `Rails.application.assets`. 

Here's the only place we access Sprockets by name:

https://github.com/reactjs/react-rails/blob/c463b8eff14806d5d4249520a7588ca1762e0f80/lib/react/rails/engine.rb#L5-L6

But I guess if you're not using Rails, you don't ever run that initializer?

sumskyi:
any chance this will be merged?

I have an app without active_record, arel, mailer, active_job

also, is `coffee-script-source` needed? probably it can be moved to the development dependencies?

rmosolgo:
I would love to merge this PR, but I'm not quite clear how the Sprockets dependency should be handled: 

- Is it _always_ required? 
- Is it only required when Rails is present? (Since the usage is in a Railtie?)

I think a good way to answer this question is either: 

- make a demo app with instructions on how to run & ensure react-rails is working properly
- add an Appraisal without Rails, or somehow add unit tests in a non-rails environment

As for coffee-script-source, perhaps we could remove it, initially it was added because of a coffee-script bug https://github.com/reactjs/react-rails/pull/168

sumskyi:
I think, the only dependencies should be there are:
* railties (ensure it is a rails app)
* **AND**
* sprockets (react-rails will not work without that)

I don't think react-rails should support non-rails environments at all, but I am talking about pretty cut rails app.

I will create a test app than.

Thanks!

rmosolgo:
@natemontgomery Thanks for your work on this! I pulled your commits into #558 along with another PR so that I could test the Sprockets dependency.

PR #430, opened 2015-12-26
Fix(SprocketsRenderer) support Sprockets::Manifest if assets were precompiled

rmosolgo:
Previously, `SprocketsRenderer` would live-compile assets, even in production. However, this is _impossible_ in Rails 5 (not just improper) because there _is no_ `Sprockets::Environment` at `Rails.application.assets`. 

Instead, there is a `Sprockets::Manifest` at `Rails.application.assets_manifest` which tells where each file's compiled counterpart is. 

So, if the Sprockets::Environment isn't available, use the manifest. 

## TODO 

- [x] Prefer the Manifest over the Environment (to avoid double-compiling assets)
- [x] Somehow, handle cases where assets are pushed to a CDN, not the local Rails server 
  - [x] ~~Check for `asset_host` config?~~, no because that doesn't necessarily mean the files are remote
  - [x] Extract `get_asset_content` to classes
  - [x] Accept a custom AssetContainer

### DIFF
--- a/lib/react/server_rendering/environment_container.rb
+++ b/lib/react/server_rendering/environment_container.rb
@@ -1,0 +1,18 @@
+module React
+  module ServerRendering
+    # Return asset contents by getting them from a Sprockets::Environment instance.
+    #
+    # This is good for Rails development but bad for production because:
+    # - It compiles the asset lazily, not ahead-of-time
+    # - Rails 5 / Sprockets 3 doesn't expose a Sprockets::Environment in production.
+    class EnvironmentContainer
+      def initialize
+        @environment = ::Rails.application.assets
+      end
+
+      def find_asset(logical_path)
+        @environment[logical_path].to_s
+      end
+    end
+  end
+end
--- a/lib/react/server_rendering/exec_js_renderer.rb
+++ b/lib/react/server_rendering/exec_js_renderer.rb
@@ -1,9 +1,9 @@
-# A bare-bones renderer for React.js + Exec.js
-# - Depends on global ReactDOMServer in the ExecJS context
-# - No Rails dependency
-# - No browser concerns
 module React
   module ServerRendering
+    # A bare-bones renderer for React.js + Exec.js
+    # - Depends on global ReactDOMServer in the ExecJS context
+    # - No Rails dependency
+    # - No browser concerns
     class ExecJSRenderer
       def initialize(options={})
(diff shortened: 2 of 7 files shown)


losvedir:
> Somehow, handle cases where assets are pushed to a CDN, not the local Rails server

I think in most cases, doesn't the CDN just serve essentially as a caching layer, with the underlying Rails server still serving the assets (the first time)?

At least in our setup with heroku/cloudfront, which I think is fairly common, we precompile assets so they're served via the `/public/assets/` folder, and we use, e.g. `<%= stylesheet_link_tag 'app' %>`. We have our `asset_host` set to our cloudfront CDN, so the generated link ends up looking like `https://abc123.cloudfront.net/assets/app-def345.css`. 

Then when the first request comes in to cloudfront after a new deploy, the filename hash will have changed, so cloudfront will proxy through to our site, and then cache and serve the asset from then on.

> * Check for asset_host config?

This wouldn't exactly work for us, I don't think. We _do_ have it set, but we still have the assets precompiled on the server as well, so maybe accessible to Sprockets? (I'm not 100% clear how sprockets works...).

Just FYI.

rmosolgo:
Yeah, I think I was off the mark on `asset_host`. It sounds like your case would be handled by the default `Sprockets::Manifest`. 

I was just poking around our own production environment, though, and I noticed that our assets _aren't_ in `public/assets`, so I'll have to figure out how to handle that!

rmosolgo:
Ug, looks like `Manifest` has changed a lot in the different sprockets versions

ssaunier:
For those who came here because of this error :

```
ActionView::Template::Error (undefined method `[]' for nil:NilClass)
```

Here's what I found. With Rails 4.2.5+, All my pre-rendered components hosted on Heroku ceased to work. I have in my `production.rb` the setting `config.assets.compile = false`. Thus `::Rails.application.assets` is always `nil` in production, so [that line](https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/sprockets_renderer.rb#L14) was the culprit.

Apparently from https://github.com/rails/sprockets-rails/issues/237 we should use the [`resolve_asset_path`](https://github.com/rails/sprockets-rails/blob/93a45b1c463a063ec7cf4d160107b67aa3db7a1a/lib/sprockets/rails/helper.rb#L74) with `allow_non_precompiled` set to `true`. I haven't dived into the specifics but from what I read it should simplify this Pull Request a lot.

Issue #433, opened 2016-01-04
read.me reference react.rb?


catmando:
Would you support a PR that added a couple of references in the readme to react.rb http://reactrb.org which is a complete ruby wrapper on react.js and uses react-rails.

rmosolgo:
Oh, sorry I missed, this! Like I said on the PR, I'm wary of putting too much in the readme about this project. 

If you want to add a "Further Reading"  / "Related Projects" / "See Also" section, you could add that project! 

Maybe it could also have a link to this: https://github.com/pboling/react-rails-benchmark_renderer

rmosolgo:
Just added to the "Related Projects" section here: https://github.com/reactjs/react-rails/pull/444

PR #435, opened 2016-01-05
added react.rb linkage

catmando:
Read.me update with suggested wording to for #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ in your Ruby on Rails (3.2+) application. `react-rails` can:
 - [Render components into views and mount them](#rendering--mounting) via view helper & `react_ujs`
 - [Render components server-side](#server-rendering) with `prerender: true`
 - [Generate components](#component-generator) with a Rails generator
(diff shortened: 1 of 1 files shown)


rmosolgo:
I don't really like this because it sends some false signals:

- it looks like the "reactjs" organization supports React.rb, but we don't have official support
- it looks like React.rb is bundled with react-rails, which it isn't. 

If another react-rails maintainer wants to support React.rb, feel free to say so. For my part, I'm busy enough with `react-rails` alone!

catmando:
@rmosolgo Thanks for considering this... would it be possible to reword this to remove your concerns?  Possibly just stating clearly that it is a separate gem, with a separate OSS team.   I really believe this is a win-win.  There are a lot of ruby folks who might try react if they could stay in ruby, and the more people using react the better off we all are.

rmosolgo:
I think it would be a good fit for a "Related Projects" list in the readme, but I don't think we should include the documentation here!

rmosolgo:
We just got a "Related Projects" section here, feel free to update the React.rb section with another PR!

PR #445, opened 2016-01-21
Update react_ujs.js.erb

maxdignan:
More idiomatic for loops

### DIFF
--- a/lib/assets/javascripts/react_ujs.js.erb
+++ b/lib/assets/javascripts/react_ujs.js.erb
@@ -43,8 +43,7 @@
     mountComponents: function(searchSelector) {
       var nodes = window.ReactRailsUJS.findDOMNodes(searchSelector);
 
-      for (var i = 0; i < nodes.length; ++i) {
-        var node = nodes[i];
+      [].forEach.call(nodes, function(node) {
         var className = node.getAttribute(window.ReactRailsUJS.CLASS_NAME_ATTR);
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
Thanks for this suggestion, can you help me understand why it is better than the existing code? Is it a style preference?

maxdignan:
It is better because it removes the instantiation of a new variable for each cycle of the loop (var node = index in nodes). Instead now, there is no variable instantiation, and instead just merely the pointer is passed into each cycle of the loops for the respective index in nodes. On top of this, does not expose an index variable, which would, to developers just entering the project, imply that some math will be done with this index variable. But in reality, this does not occur. This commit uses better practices that are better not just from performance, but also readability.

mchristen:
The performance gains are dubious at best given the usage of this code, It's mostly run once per page load. Also, if you were to look into the implementation of forEach I would imagine it's going to be something similar to what the for loop is doing manually, so you're not really saving anything, just moving the logic to a place where a developer can't see it.

I would also argue that the code is not more readable, first of all you need to understand what 'call' is doing and why there is a seemingly random empty array being used. 

I think this is a case of: "If it ain't broke, don't fix it."

rmosolgo:
> I think this is a case of: "If it ain't broke, don't fix it."

Yep, that's my instinct too. 

If it turns out that refactoring this code would improve browser support or fix a bug, let's do it, but otherwise, I think it's better to stick with what has worked well so far!

PR #446, opened 2016-01-21
add support for IE8 without jQuery

AndreaMaggi:
quick fix for IE8 without jQuery...which you probably wont/cant have with server side render


tosbourn:
:+1:

rmosolgo:
wow, ok! could you add a little explanatory comment so we don't remove it in the future?  eg `// support IE8 without jQuery`

AndreaMaggi:
Sure @rmosolgo will update it now

rmosolgo:
Cool, thanks! :tada:

PR #451, opened 2016-01-23
BabelTransformer options on a per-file basis.

danott:
### Motivation

While trying to bring some organization to client-side code we were exploring the usage of defining AMD modules, and requiring them with a tool such as [almond](https://github.com/jrburke/almond). The Babel transpiler has the ability to generate AMD modules baked in, but the we aren't able to pass the needed `moduleId` option on a per file basis.

### Solution

Provide a way to make per-file configurations by using a lambda, proc, or anything that responds to `call` and returns a `Hash` as the `jsx_transform_options`.

### Sidenote

This could be particularly useful for those who want to author their code with ES6 modules syntax. I've noticed in the issues that several users are reaching for `browserify-rails` to solve this problem. The
dependency on npm and divide between npm and sprockets leads to a lot of questions, configration, and troubleshooting. If the end-goal is to have ES6 syntax, almond and some AMD configuration will solve that problem without having to add another dependency.

### DIFF
--- a/lib/react/jsx.rb
+++ b/lib/react/jsx.rb
@@ -13,7 +13,7 @@ module React
     # You can assign `config.react.jsx_transformer_class = `
     # to provide your own transformer. It must implement:
     # - #initialize(options)
-    # - #transform(code) => new code
+    # - #transform(code, local_options = {}) => new code
     #
     # If you want to configure per-file transformation options with a proc, you may
     # accept an optional second parameter
@@ -23,13 +23,10 @@ module React
     def self.transform(code, input = {})
       self.transformer ||= transformer_class.new(transform_options)
 
-      if self.transformer.method(:transform).arity == 1
-        self.transformer.transform(code)
-      else
-        local_options = self.transform_options
-        local_options = local_options.call(input) if local_options.respond_to?(:call)
-        self.transformer.transform(code, local_options)
-      end
+      local_options = self.transform_options
+      local_options = local_options.call(input) if local_options.respond_to?(:call)
+
+      self.transformer.transform(code, local_options)
     end
   end
 end
--- a/lib/react/jsx.rb
+++ b/lib/react/jsx.rb
@@ -15,19 +15,21 @@ module React
     # - #initialize(options)
     # - #transform(code) => new code
     #
-    # If you want to configure per-file transformation options, you may
-    # optionally implement:
-    # - #transform_with_proc_options(input) => new code
+    # If you want to configure per-file transformation options with a proc, you may
+    # accept an optional second parameter
+    # - #transform(code, local_options = {}) => new code
     self.transformer_class = DEFAULT_TRANSFORMER
 
-    def self.transform(code)
+    def self.transform(code, input = {})
       self.transformer ||= transformer_class.new(transform_options)
-      self.transformer.transform(code)
-    end
 
-    def self.transform_with_proc_options(input)
-      self.transformer ||= transformer_class.new(transform_options)
-      self.transformer.transform_with_proc_options(input)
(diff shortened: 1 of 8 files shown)


rmosolgo:
Big :+1: for this feature! Weird failing test though, I'm not really sure what it even means :grimacing: . Does it do the same thing if you `rake appraisal test` locally?

danott:
@rmosolgo I think I figured it out. The test would only work in `Sprockets::VERSION == "3.0.0"` and up. `Sprockets` below that version doesn't pass anything useful, so I had to shim the test for now.

lib/react/jsx.rb
```diff
@@ -14,11 +14,22 @@ module JSX
...
+    # accept an optional second parameter
+    # - #transform(code, local_options = {}) => new code
     self.transformer_class = DEFAULT_TRANSFORMER
 
-    def self.transform(code)
+    def self.transform(code, input = {})
       self.transformer ||= transformer_class.new(transform_options)
-      self.transformer.transform(code)
+
+      if self.transformer.method(:transform).arity == 1
```
rmosolgo: How about changing `JSXTransformer#transform` to this new signature instead of checking method arity? I think the resulting code will be a bit simpler. It's _technically_ a backwards-incompatibility but I don't know of any other transformers out there! (Just `NullTransformer` in the tests :P). And the "upgrade" would be quite easy if you needed it.

rmosolgo: Maybe the new transformer API could be 

```
class Transformer 
  def initialize; end; 
  def transform(jsx_code, options); end 
end
```

Then we could remove the no-op `BabelTransformer#initialize` method below, too

lib/react/jsx/babel_transformer.rb
```diff
@@ -4,17 +4,19 @@ module JSX
     class BabelTransformer
       DEPRECATED_OPTIONS = [:harmony, :strip_types, :asset_path]
       DEFAULT_TRANSFORM_OPTIONS = { blacklist: ['spec.functionName', 'validation.react', 'strict'] }
+
       def initialize(options)
+        return unless options.is_a?(Hash)
+
         if (options.keys & DEPRECATED_OPTIONS).any?
```
rmosolgo: :+1: from me to remove this. IIRC Babel::Transpilier raises an error on invalid options anyway, so this didn't actually gain us very much!

lib/react/jsx.rb
danott: I also prefer the not-checking-arity route. With the go-ahead to make a breaking change, I'll change the implementation to what you suggested!

lib/react/jsx/babel_transformer.rb
danott: It shall be removed!

stormsilver:
I ran into this exact same problem whilst trying to cause Babel to export my files as System modules. This branch worked great with the one problem that the Sprockets 2 integration hadn't been updated to pass the correct options. I added that: https://github.com/stormsilver/react-rails/commit/018eaed20a6dd733fe6e9e86d6c9a51735b65e6e and then used the following config to cause it work:

```ruby
config.react.jsx_transform_options = ->(sprockets_input) do
  {
        modules: 'system',
        moduleIds: true,
        keepModuleIdExtensions: true,
        moduleId: sprockets_input[:name],
        filename: sprockets_input[:filename],
  }
end
```

danott:
I still think this feature would be useful. Being introspective, 9 months of me not resolving the failed tests and shepherding this to master has communicated to myself that I don't need this feature as badly as I thought I id. If someone else needs this feature and wants to pick up the code and run with it, feel free to!

PR #452, opened 2016-01-23
Add running of specs against rails 5

vipulnsward:
All seems good here.


rmosolgo:
Awesome, thanks for getting this started!

```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
An error occurred while installing activesupport (5.0.0.beta1), and Bundler
```

Looks like we'll have to only run Rails 5 tests with 2.2+.  (And .... I don't know how to do that!)

vipulnsward:
:-) Let me look at that.

vipulnsward:
@rmosolgo All's good now.

.travis.yml
```diff
@@ -1,11 +1,29 @@
...
+  - 2.2.4
   - 2.1
-  - 2.0.0
-  - 1.9.3
-  - jruby-19mode
-before_script: 'bundle exec appraisal install'
-script: 'bundle exec rake appraisal test'
+  - 2.0
+  - 1.9
+  - jruby-9.0.1.0
```
rmosolgo: Are we not supporting jruby-19mode anymore?

vipulnsward: This is 1.9 mode, and hence fails for Rails 5 right now(Specified this to pick the latest JRuby)
2.2 mode is not out yet on travis..

rmosolgo: Oh, I see, thanks!

> Ruby 1.9 mode is default in JRuby 1.7.0 and later, 1.8.7 compatibility can still be used with the --1.8 flag.

https://github.com/jruby/jruby/wiki/FAQs#does-jruby-support-ruby-19x-features

rmosolgo:
:tada: !

Issue #448, opened 2016-01-22
react-rails 2.0


rmosolgo:
There are some issues building up, I think they will need some breaking changes. Here's what's on my list. Do you have anything else that's missing?
- [x] Support Rails 5.0 
  - [x] Fix `before_filter` usage in ControllerLifecycle #404
  - [x] Fix messed up server rendering flow 
- [x] Support Sprockets 3
  -  ~~Figure out cases where assets aren't transformed  (#415, #431)~~
  - [x] Support precompiled assets (https://github.com/reactjs/react-rails/issues/443, PR:
    https://github.com/reactjs/react-rails/pull/430)
- [x] Support Sprockets 4 ?
- [x] Support Turbolinks 5
  - Stop warning about Turbolinks 2.2.4 
  - Update for new Turbolinks API 
  - Split UJS into different files, make `react_ujs` a manifest?
- [x] Reduce Rails dependency 
  - [x] Altogether here: #558 
  - ~~Basically, just merge this: https://github.com/reactjs/react-rails/pull/427~~
  - ~~Don't depend on sprockets #493~~ 
-  [x] Improve the flexibility of React addons 
  - use `webpacker`
  - ~~Support custom ReactJS builds (backed with npm & webpack, built into the apps asset directory)~~
  - ~~Expose `renderToString` etc without `//= require react-server` (https://github.com/reactjs/react-rails/issues/442)~~
- [x] General cleanup 
  - [x] Use `.js` files for CONSOLE_POLYFILL and CONSOLE_REPLAY
  - [x] Can we merge `engine.rb` and `railtie.rb`? They're almost the same thing, right?
  - [x] Fix replay_console + server rendering #555 
-  [x] Support `require()` environments #678 
  - Expose UJS 
  - Document custom component-fetching function 
- [x] Improve documentation 
  - ~~[ ] Improve YARD docs~~
  - ~~[ ] Split readme into guides & improve content~~
  - [x] Update readme

twe4ked:
Thanks for creating this issue @rmosolgo. Looks like #450 could be a dot-point under _Support Rails 5.0_.

rmosolgo:
Thanks, I added a note!

SylarRuby:
I've seen this list for version 2. Ok, support for rails 5 as I'm getting this: https://github.com/reactjs/react-rails/issues/467

connorshea:
[...]

PR #455, opened 2016-01-26
Use .js files for CONSOLE_POLYFILL and CONSOLE_REPLAY

danott:
One small PR to begin chipping away at 2.0.0! #448


rmosolgo:
:moneybag:  I think that makes the file much clearer! Thanks!

PR #456, opened 2016-01-26
Support controller callbacks for Rails 3..5

### LINKED ISSUE #404: Depreciation warnings with Rails 5
react-rails is using `before_filter`, which is deprecated in Rails 5 and will disappear in 5.1 : 

```
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead. (called from block in <module:ControllerLifecycle> at gems/react-rails-290765a1ac14/lib/react/rails/controller_lifecycle.rb:8)
DEPRECATION WARNING: after_filter is deprecated and will be removed in Rails 5.1. Use after_action instead. (called from block in <module:ControllerLifecycle> at gems/react-rails-290765a1ac14/lib/react/rails/controller_lifecycle.rb:9)
```

`before_filter` has been renamed to `before_action` in Rails 4. I dont know what your policy is regarding deprecations like this on not-yet released versions, but I think it might be worth it to switch to the new name and have a compatibility alias for Rails 3.2.

danott:
`before_filter` is deprecated in Rails 5 and will disappear in 5.1

This change begins using `before_action`, with a fallback to `before_filter` if `before_action` is not available.

closes #404

### DIFF
--- a/lib/react/rails/controller_lifecycle.rb
+++ b/lib/react/rails/controller_lifecycle.rb
@@ -4,9 +4,11 @@ module React
       extend ActiveSupport::Concern
 
       included do
-        # use old names to support Rails 3
-        before_filter :setup_react_component_helper
-        after_filter :teardown_react_component_helper
+        # use both names to support Rails 3..5
+        before_action_with_fallback = begin method(:before_action); rescue method(:before_filter); end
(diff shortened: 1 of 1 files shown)


rmosolgo:
:moneybag: :fireworks: Thanks!

PR #461, opened 2016-01-31
Remove data-react-props attribute after mounting component

bluesh55:
This feature makes not to expose props data to html by removing 'data-react-props' attribute after mounting component.

I think it is good idea better than to expose all props data.


rmosolgo:
Currently, mounting components into the page is idempotent. It can be done over and over with no problems. But this change makes it destructive: once you mount a component, you can't remount it (because its props have been lost). That's why the test is failing. 

Could you help me understand why this change is desirable?

Could you handle this at app-level if it wasn't added to `react-rails`? Or maybe you could make this opt-in instead of default.

bluesh55:
Oh.. i didn't think that point.

it's complex problem than i think.

thanks rmosolgo.

PR #463, opened 2016-02-02
#462 Fixed updating application.js whith messed-up require markers

### LINKED ISSUE #462: Installation fails if you don't have require_tree in application.js
Hey there!

I don't know what's the accepted practice for injection point detection, but it's present version fails (https://github.com/reactjs/react-rails/blob/4c8585772c02df93ecc32ae989b0a04f70b57dff/lib/generators/react/install_generator.rb#L28 fails with "undefined method [] for nil") when you either don't have require_tree or have it commented (like "// require_tree .") in application.js

If you have an idea of how to fix this, let me know, I'll send you PR. Otherwise, I'll give it some thought later.

alg:
Updated checks to be totally in sync and also added handling for extra spaces in every possible place. Fixes #462

### DIFF
--- a/lib/generators/react/install_generator.rb
+++ b/lib/generators/react/install_generator.rb
@@ -22,11 +22,10 @@ module React
         if manifest.exist?
           manifest_contents = File.read(manifest)
 
-          if manifest_contents.include? 'require turbolinks'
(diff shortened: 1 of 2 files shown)


rmosolgo:
nice, thanks!  :confetti_ball:

PR #466, opened 2016-02-03
add ability to pass in a custom selector function to mountComponents

rearadmiral:
* this is to support the case where you want to mount or unmount a single component
  with a known ID that is a sibling of one or more components that you don't want to affect
* currently, if you pass in a parent element to use a scope, it will mount/unmount
  all components underneath
* you also can't pass in the known ID of a mounted component because the selector
  it constructs assumes the react element is a child

The context for this change is we have a custom view helper that goes through a bunch of hoops to properly render React components used in older Rails pages that replace portions of their page with ajax and the Protoype/jQuery helpers. We need a way to mount components that are inserted into the page this way and we have the element's IDs. Unfortunately, none of the existing parameter options in mountComponents will allow us to mount a single element with a known ID.

We're currently using this hack to make it work:
`ReactRailsUJS.mountComponents('#knownID, html >');`

Due to the way it builds the selector string, this purposely creates a non-matching selector for the second bit, leaving only the ID selector to match. But this is britle and relies on the impl of ReactRailsUJS not changing. I'd love to get rid of this hack and use this impl instead.

### DIFF
--- a/lib/assets/javascripts/react_ujs_mount.js
+++ b/lib/assets/javascripts/react_ujs_mount.js
@@ -27,7 +27,8 @@
           parent = searchSelector;
           break;
         case 'string':
-          selector = searchSelector + ' [' + window.ReactRailsUJS.CLASS_NAME_ATTR + ']';
+          selector = searchSelector + '[' + window.ReactRailsUJS.CLASS_NAME_ATTR + '], ' +
+                     searchSelector + ' [' + window.ReactRailsUJS.CLASS_NAME_ATTR + ']';
           parent = document;
           break
         default:
--- a/test/dummy/app/views/pages/show.html.erb
+++ b/test/dummy/app/views/pages/show.html.erb
@@ -4,11 +4,14 @@
 </ul>
 
 <div id='component-parent'>
-  <%= react_component 'HelloMessage', :name => @name %>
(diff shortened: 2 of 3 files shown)


rmosolgo:
It seems important to support the case you're describing! I'm surprised it hasn't come up already.

What if `findDOMNodes` searched the matching node AND its children? I think that would work the same for existing users and it would handle this case nicely. (Also, it wouldn't expand the API of this function!)

rearadmiral:
That would be preferable for me as well. I was just worried it would break existing use cases but I guess that's actually unlikely. I'll revise this PR.

rearadmiral:
So, modifying the string case option to create a selector string like this:

"#MySelector[data-react-class], #MySelector [data-react-class]"

Was that along the lines of what were thinking?

rmosolgo:
Yeah, that makes sense! if the given selector is a react component, it will unmount that, or it will unmount any child react components. 

I guess technically you could have a bad time if any children of the react component have `data-react-class` but ... I don't think anyone would do that!

rearadmiral:
Overwrote the branch with a new implementation. Let me know if this seems good.

rmosolgo:
That looks great! Those failures look totally unrelated:

image


Maybe I'll have to find another way to get a recent `phantomjs` install for Travis!

rmosolgo:
:ship: in 1.6.1, thanks!

Issue #443, opened 2016-01-16
Server-side rendering is not working with sprockets-rails 3.0


oleglitvin:
In sprockets-rails 3.0 `Rails.application.assets` is disabled when `config.assets.compile=false` (https://github.com/rails/sprockets-rails/pull/220).
And `React::ServerRendering::SprocketsRenderer` is failing on production environment at [this line](https://github.com/reactjs/react-rails/blob/122ae3e0f5595254db0b292e19c9994ebc1f069c/lib/react/server_rendering/sprockets_renderer.rb#L14), with `undefined method '[]' for nil:NilClass`

oleglitvin:
OK, there is an opened pull-request https://github.com/reactjs/react-rails/pull/430/files

rmosolgo:
Could you give that branch a try? Let me know if it works for you -- it looks good to me but I'd like a second opinion before I merge it!

inchr:
I need this too :( I will test your PR very soon @rmosolgo ! Thanks for your work!!

DTwigs:
This would be super helpful!

oleglitvin:
> Could you give that branch a try? Let me know if it works for you -- it looks good to me but I'd like a second opinion before I merge it!

@rmosolgo yep, it works!

oleglitvin:
It fails at this line https://github.com/reactjs/react-rails/pull/430/files#diff-543f6fa05678c006b8d4baa5cd4b517cR13, not being able to find `react-server.js` (even if react-server was added to the `config.assets.precompile` and being precompiled)

djforth:
+1

dbackeus:
+1

rmosolgo:
Could you try react-rails 1.6.0? It includes #430 which was my first try at fixing this

tirdadc:
I can confirm that I have this issue in production with 1.6.0 when server-side rendering is used.

`ActionView::Template::Error (No compiled asset for react-server.js, was it precompiled?):`

rmosolgo:
Ok, thanks, let me try to track that down!

rmosolgo:
Ohh strange. When it creates the manifest, the manifest has `production/react-server` instead of `react-server`. Then, SprocketsRenderer fails to find `react-server` since it has a different logical path. 

I found a workaround: 

1. Make sure components.js is compiled (`Rails.application.config.assets.precompile += %w( components.js )`)
1. Add `//= require react-server` to your components.js file 
[...]

PR #471, opened 2016-02-08
refactor(ServerRendering) simpler server rendering assets

rmosolgo:
We have a problem with Sprockets 3. 
- We add `lib/assets/react-source/{ production | development }` to the asset pipeline 
- This makes it so `//= require react` works for production build or development build 
- For some reason, Sprockets 3 puts those files in the manifest as `production/react-server` instead of `react-server` (https://github.com/reactjs/react-rails/issues/443#issuecomment-180544359)
- As a result, we can't find the file from the manifest. (The `logical_path` has changed!)

Why does it do that?! I don't know. But I think this is a chance to simplify the default server rendering setup. 

Right now, we require `["react-server.js", "components.js"]`. Because `react-server` is "automatic", it's a bit hidden. I think we should add a _new_ file called `server_rendering.js` which includes those two files. This is good because: 
- By default, you have total control over the SSR manifest 
- Less "automatic" stuff that can just break and stop you from building a sweet app 
- `react-server.js` becomes a public API (I'm using this for my app already, I need `renderToString`)

However, this introduces some breaking changes: 
- To upgrade, you should add the `server_rendering.js` manifest 

**TODO:**
- Why are the compiled assets different?


franzejr:
I agree README is not totally friendly related to the react-server file.

caseys:
"automatic" react-server.js has been confusing for me too.

PR #475, opened 2016-02-11
Refactor UJS, support Turbolinks 5

rmosolgo:
Turbolinks 5 might have some different APIs to support. We want to support all of them!

I think this a good chance to split the UJS into different files. I hope it will make it easier for new people to find the code they want to change & work on it. Also, you could "mix & match" the UJS if you didn't want certain parts of it. 

- [x] Refactor UJS into responsibility-specific files 
- [x] Support Turbolinks 2.4.0+ only (a smaller version will throw an error, you could work around it by assigning `Turbolinks.EVENTS` yourself)
- [x] Add gemfiles specifying Turbolinks 5 / Turbolinks Classic
- [x] Detect & support Turbolinks 5
- [x] Install a recent version of PhantomJS so that the tests _actually_ test Turbolinks!
- [x] Force specific turbolinks versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,9 @@ matrix:
       rvm: 2.1
   allow_failures:
     - rvm: jruby-9.0.1.0
+
+before_install:
+  - mkdir travis-phantomjs
+  - wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2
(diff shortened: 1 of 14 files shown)


rmosolgo:
Maybe the old one will still work thanks to 

https://github.com/turbolinks/turbolinks/blob/master/src/turbolinks/compatibility.coffee

rmosolgo:
- [Rails 4.2 + Turbolinks 2.5](https://travis-ci.org/reactjs/react-rails/jobs/108691229)
  
  image

- [Rails 5.0 + Turbolinks 5](https://travis-ci.org/reactjs/react-rails/jobs/108691231)

  image

  image

:tada: :question:

rmosolgo:
How could this possibly work? Turbolinks 5 doesn't have `Turbolinks.EVENTS`!

rmosolgo:
That's more like it: 

image

PR #453, opened 2016-01-24
feat(Sprockets) add sprockets 2 and 3 to Appraisals

rmosolgo:
There are some known bugs with Sprockets incompatibilities and I'd like to make those more reproducible (and maintain compatibility with Sprockets 2)

PR #478, opened 2016-02-19
Use Manifest#dir for sprockets 3 compatibility

dudeman:
This changed in sprockets a while ago https://github.com/rails/sprockets/commit/3255c8ce8623d54dcf1d7676c45963da492237cd
But it's nice to support those using the ancient 3.X version.

### DIFF
--- a/lib/react/server_rendering/manifest_container.rb
+++ b/lib/react/server_rendering/manifest_container.rb
@@ -11,7 +11,7 @@ module React
 
       def find_asset(logical_path)
         asset_path = @manifest.assets[logical_path] || raise("No compiled asset for #{logical_path}, was it precompiled?")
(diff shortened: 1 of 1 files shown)


rmosolgo:
This seems good! 

But now I'm trying to figure out ... how did this _ever_ work!?

dudeman:
Yeah, good question. I'm not sure.  But somehow it broke for us when we updated from 1.5.0 to 1.6.0

rmosolgo:
I cherry-picked it to #453  to make sure it was good on all versions of Sprockets. And it looks like it is! So, it got merged to master along with that.

Thanks for this fix!

rmosolgo:
:ship: in 1.6.1, thanks!

PR #485, opened 2016-02-27
Add YamlManifestContainer for Rails 3.2 server rendering support

### LINKED ISSUE #483: Let's talk about Rails 3.2 -- Problem with precompiled assets and server rendering
Hi everyone!

Sorry if this comes across as a silly question, but I'm seeing mixed messages about Rails 3.2 support and want to get a definitive answer. Based on my limited testing it seems nearly everything in Rails 3.2 works _except_ server rendering (ala `prerender: true`) with precompiled assets. I was a bit surprised to find out that there was a known issue with loading the manifest file in 3.2 (see [this comment](https://github.com/reactjs/react-rails/blob/master/test/react/server_rendering/manifest_container_test.rb#L3-L4) for more details). This all started when I got a `undefined method 'assets_manifest'` error In production and started digging into the source. 

As far as I can tell there isn't anything listed in the README about this limitation and I'm just looking to get a sense of how well supported Rails 3.2 is by the project's maintainers. I realize 3.2 is nearing its end of life, so perhaps this is a good time to reevaluate its support. I'd hate for someone else to get as deeply invested in using this gem, like I did, only to find out one of its core features doesn't work in production. :disappointed: 

That all being said I'd love to come up with a solution to this issue. I'm not terribly familiar with the internals of the asset pipeline but if there is some way to get this feature working in 3.2 I don't mind rolling up my sleeves and getting it done with a little guidance.

Thanks for all of your awesome work. I really appreciate it :sunflower:

jordanbyron:
Based on discussions with @rmosolgo this PR restores full support for Rails 3.2, specifically server rendering with precompiled assets.

Closes #483 and Closes #484 

Added tests pass but `react_test/precompiling assets works` throws an error when trying to call `clear_precompiled_assets`

### DIFF
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -1,5 +1,7 @@
 require "react/server_rendering/environment_container"
 require "react/server_rendering/manifest_container"
+require "react/server_rendering/yaml_manifest_container"
+
 module React
   module ServerRendering
     # Extends ExecJSRenderer for the Rails environment
@@ -62,7 +64,11 @@ module React
         elsif ::Rails.application.config.assets.compile
           EnvironmentContainer.new
         else
-          ManifestContainer.new
+          if Rails::VERSION::MAJOR == 3
+            YamlManifestContainer.new
+          else
+            ManifestContainer.new
+          end
         end
       end
     end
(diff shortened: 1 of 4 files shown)


rmosolgo:
Good call to just precompile assets with 3.2 to make sure it works, that's way better than my suggestion to copy random files!

Could you switch `clear_precompiled_assets` to use `rm_rf` here? 

https://github.com/reactjs/react-rails/blob/master/test/test_helper.rb#L51

I think that would make it skip the directory if it doesn't exist.

jordanbyron:
@rmosolgo done and the 3.2 tests are passing on my machine. Hopefully Travis agrees! :dizzy_face:

rmosolgo:
:moneybag: :rocket: Thanks so much! that's great.

rmosolgo:
:ship: in 1.6.1, thanks!

PR #487, opened 2016-02-29
Fix Rails reference in SprocketsRenderer

jordanbyron:
Hi @rmosolgo!

Unfortunately I found an issue with [the code we just merged](https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/sprockets_renderer.rb#L67) that checks the Rails version in `SprocketsRenderer`. Rather than reference the top level `Rails` class using `::Rails` I mistakenly referenced plain old `Rails` which in this context is `React::Rails`. I expected that to be tested but on second glance saw it was not. This PR fixes that reference plus adds test coverage for `SprocketsRenderer` when `Rails.application.config.assets.compile == false`.

I've also fixed a few issues I found with the test helper's `precompile_assets` method. Let me know what you think. :dog:

### DIFF
--- a/test/react/server_rendering/sprockets_renderer_test.rb
+++ b/test/react/server_rendering/sprockets_renderer_test.rb
@@ -57,6 +57,9 @@ class SprocketsRendererTest < ActiveSupport::TestCase
 
   test '#render returns html when config.assets.compile is false' do
     begin
+      Rails.application.config.assets.precompile += [
+        "react-server.js", "components.js"]
+
       precompile_assets
 
       Rails.application.config.assets.compile = false
--- a/test/react/server_rendering/sprockets_renderer_test.rb
+++ b/test/react/server_rendering/sprockets_renderer_test.rb
@@ -54,4 +54,22 @@ class SprocketsRendererTest < ActiveSupport::TestCase
     end
     assert_match(/ReferenceError/, err.to_s, "it doesnt load other files")
   end
+
+  test '#render returns html when config.assets.compile is false' do
+    begin
+      precompile_assets
+
+      Rails.application.config.assets.compile = false
+
+      @renderer = React::ServerRendering::SprocketsRenderer.new({})
+
+      result = @renderer.render("Todo", {todo: "write tests"}, nil)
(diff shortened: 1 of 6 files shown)


test/react/server_rendering/sprockets_renderer_test.rb
```diff
@@ -54,4 +54,25 @@ class SprocketsRendererTest < ActiveSupport::TestCase
     end
     assert_match(/ReferenceError/, err.to_s, "it doesnt load other files")
   end
+
+  test '#render returns html when config.assets.compile is false' do
```
rmosolgo: :moneybag:

test/test_helper.rb
```diff
@@ -40,18 +40,56 @@ def asset.fresh?(env); false; end
...
+  if Rails::VERSION::MAJOR > 3
+    # Make a new manifest since assets weren't compiled before
+    config = Rails.application.config
+    path = File.join(config.paths['public'].first, config.assets.prefix)
+    new_manifest = Sprockets::Manifest.new(Rails.application.assets, path)
+    Rails.application.assets_manifest = new_manifest
+  end
+
+  assets_directory = File.expand_path("../dummy/public/assets", __FILE__)
+  raise "Asset precompilation failed" unless Dir.exists?(assets_directory)
```
rmosolgo: oh, was that the problem before? I kind of assumed it was test order-dependent, like some test wasn't cleaning up after itself

jordanbyron: Yes. Before I added a test for the `YamlManifestContainer` class assets were only precompiled once which worked, but as soon as the code attempted to precompile them a second time it silently failed (because `assets:precompile:all` et al were not `reenabled`) hence the problem with `clear_precompiled_assets`.

I added the raise here to be more explicit with our expectations.

rmosolgo:
Thanks again not just for the quick fix but also adding the test for the future!

jordanbyron:
No problem at all. I should have verified the new code I added had test coverage so it's the least I can do. Thanks again for all your help and for doing an awesome job maintaining this project :hamster:

rmosolgo:
shipped in 1.6.2

PR #490, opened 2016-03-07
added pjax support

QuantumKing:


rmosolgo:
Rebased the first 3 commits here:  #518

The last one isn't needed in React 15+ :D 

Thanks for this feature!

PR #493, opened 2016-03-09
Register with sprockets when sprockets is available

guzart:
This allow to use react-rails without sprockets. User can use their own renderer without the need for sprockets.

This was a quick proof of concept, please let me know what would you like the style code to be.

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -59,8 +59,10 @@ module React
           addons: app.config.react.addons,
         })
 
-        sprockets_env = app.assets || app.config.assets # sprockets-rails 3.x attaches this at a different config
-        sprockets_env.version = [sprockets_env.version, "react-#{asset_variant.react_build}",].compact.join('-')
+        sprockets_env = app.assets || app.config.try(:assets) # sprockets-rails 3.x attaches this at a different config
+        if !sprockets_env.nil?
+          sprockets_env.version = [sprockets_env.version, "react-#{asset_variant.react_build}",].compact.join('-')
+        end
 
       end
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
I like this idea, I think it would go well with #427 ! Your code style is fine by me.

One question: how can we be sure to keep maintaining this in the future? Can we test somehow that react-rails still works in environment _without_ sprockets?

guzart:
I did a quick google search on how to test "railties", it seems like it's considered an integration test with Rails. Looking for examples I foudn that the Draper gem has an actual [Rails app fixture](https://github.com/drapergem/draper/tree/master/spec/dummy/app). 

I'll keep looking and thinking on how to test the conditional integration.

rmosolgo:
Yeah, I was wondering if we might be able to add an appraisal with some configuration that didn't have sprockets ... but, sprockets is a dependency of Rails (right?) so I'm not sure how customizing the gemfile could do that :S

rmosolgo:
Just curious, how do you run Rails _without_ Sprockets? Isn't Sprockets a dependency of Rails (via `sprockets-rails`?

guzart:
I guess you can't help installing the gem since it's a direct dependency of the Rails gemspec. 

TLDR; you remove the `require "sprockets/railtie"` line from your application.rb file.

When generating a site you can add the `--skip-sprockets` flag which generates an `application.rb` file with the list dependencies instead of just `require "rails/all"`, and the `require "sprockets/railtie"` line is [omitted](https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/railties/lib/rails/generators/rails/app/templates/config/application.rb#L15).

rmosolgo:
@guzart Thanks for your work on this! I pulled your commit into #555 so I could rebase it & set up a test run without the Sprockets railtie.

Issue #494, opened 2016-03-11
Console Replay and Line Feeds


wied03:
I haven't tested this outside of Chrome, but i've had better luck with console replay working if I eliminate the 2 backslashes below. Is this a remnant of when this was not in its own JS file (was a Ruby string before??)

This is working OK for me.

```js
(function (history) {
  if (history && history.length > 0) {
    result += '\n<scr'+'ipt>';
    history.forEach(function (msg) {
      result += '\nconsole.' + msg.level + '.apply(console, ' + JSON.stringify(msg.arguments) + ');';
    });
    result += '\n</scr'+'ipt>';
  }
})(console.history);
```

rmosolgo:
> Is this a remnant

Yes, I bet it is! would you be able to submit a patch?

wied03:
I'll see if I can do that.

PR #496, opened 2016-03-12
Fix bug with console replay and carriage returns

wied03:
Code from Ruby string days was still in the standalone JS file and it caused browsers to complain. Tightened up the regex to catch this in tests.

Fixes https://github.com/reactjs/react-rails/issues/494


wied03:
@rmosolgo - Are the travis/other hooks manual or did github not run them?

rmosolgo:
Hmmm... it supposed to run automatically, it did for some previous PRs! https://travis-ci.org/reactjs/react-rails/requests

wied03:
I toggled it. Maybe that will work.

rmosolgo:
:moneybag:  thanks for the fix! nice find and thanks for including a test.

PR #497, opened 2016-03-13
When a static render is done, don't include react class/props

wied03:
Should result in cleaner markup in cases where the client side React won't be rendered anyways.

Alternatively, this could be an option but it seems like a sensible default to me.

### DIFF
--- a/lib/react/rails/component_mount.rb
+++ b/lib/react/rails/component_mount.rb
@@ -32,9 +32,11 @@ module React
         end
 
         html_options = options.reverse_merge(:data => {})
-        html_options[:data].tap do |data|
-          data[:react_class] = name
-          data[:react_props] = (props.is_a?(String) ? props : props.to_json)
+        unless prerender_options == :static
+          html_options[:data].tap do |data|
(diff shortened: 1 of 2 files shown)


rmosolgo:
What if you want the server to render the _initial_ html, but you want a dynamic UI when the page loads? 

Currently, the browser will pick up where the server left off, using those props. 

Happy to include this behavior as opt-in, but I don't want to change the default without feedback from other server-side rendering users! (I am not one.)

wied03:
If you're have the server do the initial render, which I often do, you won't do a static render, you would just do a prerender: true. Otherwise I think there would be little point to a server side render because without checksums, etc, the client side has to do everything again.

rmosolgo:
Ohh I see, I misunderstood! 

I agree that this is a safe assumption. It's technically a breaking change but ... only if you're using the library really inefficiently!

rmosolgo:
(maybe I can let this issue rest here in case any other watchers have a reason _not_ to merge it!)

rmosolgo:
Thanks!

PR #503, opened 2016-03-24
Update to accept new babel 6 module exports

dphaener:
With the update to Babel 6, modules are being exported differently. For
instance, the default module is now under a default key. This is
breaking things when using an expose loader with webpack. This adds a
check for that before trying to import the module.

### DIFF
--- a/lib/assets/javascripts/react_ujs_mount.js
+++ b/lib/assets/javascripts/react_ujs_mount.js
@@ -16,7 +16,7 @@
 
     // If that didn't work, try eval
     if (!constructor) {
-      constructor = eval.call(window, className)'
+      constructor = eval.call(window, className);
     }
 
     // Lastly, if there is a default attribute try that
--- a/lib/assets/javascripts/react_ujs_mount.js
+++ b/lib/assets/javascripts/react_ujs_mount.js
@@ -2,6 +2,31 @@
   // jQuery is optional. Use it to support legacy browsers.
   var $ = (typeof window.jQuery !== 'undefined') && window.jQuery;
 
+  // Get the constructor for a className
+  // Not a part of the public API
+  function getConstructor(className) {
+    // Assume className is simple and can be found at top-level (window).
+    // Fallback to eval to handle cases like 'My.React.ComponentName'.
+    // Also, try to gracefully import Babel 6 style default exports
+    //
+    var constructor;
+
+    // Try to access the class globally first
+    constructor = window[className];
+
+    // If that didn't work, try eval
+    if (!constructor) {
+      constructor = eval.call(window, className)'
+    }
+
+    // Lastly, if there is a default attribute try that
+    if (constructor && constructor.default) {
+      constructor = constructor.default;
+    }
+
+    return constructor;
+  }
+
   window.ReactRailsUJS = {
     // This attribute holds the name of component which should be mounted
     // example: `data-react-class="MyApp.Items.EditForm"`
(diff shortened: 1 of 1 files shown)


rmosolgo:
Interesting, could you share a couple files that have the issue? Like, one file that defines a component and another that puts the component in the global namespace? 

I wonder if, to support this case, it could be: 

```js 
var constructor
// First, try accessing the name globally: 
constructor = window[className]

// If that doesn't work, try eval (to handle "My.Namespaced.Component")
if (!constructor) { 
 constructor = eval.call(window, className)
}

// Then, if it has a `.default` attribute, we assume it's exported from Weback in that way: 
if (constructor && constructor.default) {
  constructor = constructor.default 
}
```

lib/assets/javascripts/react_ujs_mount.js
```diff
@@ -53,7 +53,7 @@
 
         // Assume className is simple and can be found at top-level (window).
         // Fallback to eval to handle cases like 'My.React.ComponentName'.
-        var constructor = window[className] || eval.call(window, className);
+        var constructor = window[className].default || window[className] || eval.call(window, className);
```
rmosolgo: if `window[className]` returns undefined, will this throw `TypeError: Cannot read property 'default' of undefined` ?

dphaener: Yup, it sure would. Sorry about the delay on this. I think the solution you outlined above would work just fine. I'll update and resubmit.

dphaener:
@rmosolgo Cleaned this up a bit and rebased current master. Hopefully this will pass CI. 👍

lib/assets/javascripts/react_ujs_mount.js
```diff
@@ -2,6 +2,31 @@
...
+    // Also, try to gracefully import Babel 6 style default exports
+    //
+    var constructor;
+
+    // Try to access the class globally first
+    constructor = window[className];
+
+    // If that didn't work, try eval
+    if (!constructor) {
+      constructor = eval.call(window, className)'
```
rmosolgo: oopsie `'`

dphaener:
And to answer your other question @rmosolgo I have bundle files that look something like this:

```js
require('expose?React!react');
require('expose?Navigation!../javascripts/application/views/shared/navigation');
require('expose?Flash!../javascripts/application/react-components/flash');
require("font-awesome-webpack!../stylesheets/font-awesome.config.js");
require('../stylesheets/bundles/breweries');
```

Then it's being bundled up by webpack with that file as the entry point. I'm exposing, in this instance, the Navigation and Flash components globally, but they get exported with the `default` object. This is the new Babel behavior. I tried many things to get it to work without using this workaround, but short of going back to Babel 5 I haven't been able to figure out anything.

lib/assets/javascripts/react_ujs_mount.js
```diff
@@ -2,6 +2,31 @@
   // jQuery is optional. Use it to support legacy browsers.
   var $ = (typeof window.jQuery !== 'undefined') && window.jQuery;
 
+  // Get the constructor for a className
+  // Not a part of the public API
+  function getConstructor(className) {
```
rmosolgo: this function could be _really_ useful for people using NodeJS-type builds, could you make it part of `ReactRailsUJS`? 

(That way, people could override `getConstructor` to hook into their build systems!!)

dphaener: Oh I didn't think about that. Great idea! Coming...

rmosolgo: Am I reading the diff right? It looks like this function is in the source twice, but we only need the one that's attached to ReactRailsUJS, right?

dphaener:
@rmosolgo Ok then. Guess I missed that. Whoops.

rmosolgo:
Wow, sorry I left this sitting around for so long! I was just thinking "I should write some docs for that new component lookup function" and then I didn't find it on master 😬 

Thanks for this!

rmosolgo:
Shipped in 1.8.0!

dphaener:
:shipit:  👍

PR #508, opened 2016-03-31
Updated README to show latest gem version

jasonschock:


rmosolgo:
I _think_ `gem "react-rails", "~> 1.6.0"` is actually a pretty good instruction -- it will install the latest patch version but won't increase the minor version, right?

Really, I don't want this to become a maintenance bug. Eventually, we'll just forget to update it and people will copy-paste the wrong thing 😖 Could we remove it altogether? Maybe the gem badge at the top header is enough?

jasonschock:
Sure, good points. I think I'd vote to remove it altogether then.

rmosolgo:
If you'd like to remove it, feel free to open another PR or reopen this one!

PR #512, opened 2016-04-13
React 15.0.1

jakegavin:
[React v15.0](https://facebook.github.io/react/blog/2016/04/07/react-v15.html) and [React v15.0.1](https://facebook.github.io/react/blog/2016/04/08/react-v15.0.1.html) were released last week. It'd be awesome to get react-rails onto v15.0.1 as well. 

Two things to note that I wasn't quite sure about: 
-  The previously deprecated `cloneWithProps` addon has been removed. In this PR, I have removed it from `lib/assets/react-source/development`, `lib/assets/react-source/production`, and `react-builds/`. Should we have done deprecation warnings on the gem side of things or were the React warnings sufficient? I believe React warnings were sufficient but figured I'd double check. 
- `React.render` and `React.unmountComponentAtNode` were removed so I removed the fall backs to those in the `react_ujs_mount`. I grepped and didn't see any other references to removed deprecations. I'm not quite sure why we needed to have those fallbacks when the gem had React v0.14 so I figured I'd double check that it was okay to remove them.

Thanks for all the work you put into this gem. Let me know if theres anything I should change about this PR!


jnstq:
It would be nice if this got merged

starsirius:
+1

rmosolgo:
Sorry for the delay! Thanks for bumping it. 

Emoji responses are great but they don't trigger a notification ... I wonder if there's a setting for that. 

Anyways, yeah, I think the previous React deprecations are enough for our case!

jnstq:
We run in to this issue, https://github.com/facebook/react/issues/6547 our select with optgroup stopped working. I guess it will be fixed in next point release.

PR #518, opened 2016-04-26
Pjax support 2

rmosolgo:
Thanks to @QuantumKing for actually doing the work on this!


rmosolgo:
They all passed until we got a 403 for PhantomJS ... I'm calling it good :P (And trying to fix it in #519)

PR #519, opened 2016-04-26
fix(build): use alternate url for phantomjs to avoid 403 throttling errors

rmosolgo:
The old endpoint is throttled


rmosolgo:
omg it finally worked 😩

PR #522, opened 2016-04-28
Use register_engine on Sprockets 3

mikesea:
Hi! I'm upgrading a project that uses react-rails from Sprockets 2.x to 3.x. When precompiling assets, it appeared as though we were precompiling every single `.jsx` file, causing asset precompilation to take a very long time.

After digging around (and originally believing the issue was in Sprockets itself: https://github.com/rails/sprockets/issues/289), I discovered the issue was because of how react-rails registers the transformers/processors for handling `.jsx` files in Sprockets 3. What is currently in place should work fine for Sprockets 4 (still unreleased), but in Sprockets 3 it causes any custom mime type to be added to the list of precompiled assets in addition to being merged into the bundled asset.

This change just uses the `register_engine` method to handle `.jsx` transformations, which fixes the issue for Sprockets 3. This matches how `.coffee` files are handled in Sprockets 3 as well: https://github.com/rails/sprockets/blob/3.x/lib/sprockets.rb#L125


rmosolgo:
Thanks for this! I wonder if it will fix some of the mysterious non-transforming assets with Sprockets 3 too.

In general, it looks good to me. Looking back at the PR where it was added (https://github.com/reactjs/react-rails/pull/385), I _think_ it was just to future-proof for Sprockets 4. But maybe we moved too soon! 

Regarding Sprockets 4 support, you could either: 

- Remove it, since it's untested (and currently unsupported) 
- Add Sprockets 4 support: 
  - Add a gemfile to `gemfiles/` which includes Sprockets 4 
  - Add the gemfile to the list in `.travis.yml` (and add an entry to the `exclude:` for Ruby 2.1)
  - Make sure the tests pass 

Your call!

mikesea:
So, tests fail when adding a Sprockets 4 gemfile, which seems to be due to a few things:

- Missing the new `app/assets/config/manifest.js` file in the dummy app. 
- Chaining processors (e.g., `.coffee` -> `.jsx` -> `.js`) seems to have changed in Sprockets 4. I'm not even sure it's possible to support coffeescript the same way it's done now. For example, Sprockets 4 will write the `components/Todo.js.jsx.coffee` file to `components/Todo.js.jsx.js`. At that point, the file still needs to be JSX transformed. I tried a few things (like registering another mime-type to process these files through the JSX processor), but nothing seemed to work.
- `ReactTest#test_the_development_version_with_addons_is_loaded:
NoMethodError: undefined method 'pathname' for #<Sprockets::Asset:0x007fd3390b3c68>`
- A few of the server rendering tests fail as well.

That said, using `react-rails` as it exists today on Sprockets 4 does work in some cases. I'm hesitant to drop support entirely, especially for anyone that might be getting by, but there's a good amount of work necessary to make all of these tests pass.

mikesea:
How about this: I'll add the gemfiles for sprockets 4, update `.travis.yml` and allow failures only for the sprockets 4 builds, and add a note to the readme that says Sprockets 4 support is "use at your own risk" for the time being. I'd love to help with Sprockets 4 support, but I think it's okay to punt on that for the moment.

rmosolgo:
> update .travis.yml and allow failures

🎉  💰  😹  sounds perfect!

mikesea:
@rmosolgo done!

mikesea:
@rmosolgo Anything else I can do to get this merged?

rmosolgo:
Just remind me, like you did :) Sorry about the delay, and thanks for the fix!

mikesea:
Thank you!

rmosolgo:
:ship: shipped in 1.7.1!

PR #526, opened 2016-05-05
Fix Rails 5 .to_prepare deprecation warning

molawson:
### Changes

In rails/rails@d3c9d808e3e242155a44fd2a89ef272cfade8fe8, `ActionDispatch::Reloader.to_prepare` (and friends) were moved to `ActiveSupport`.

For now, they're only issuing a deprecation warning, but they're planning on dropping support in Rails 5.1.

> `DEPRECATION WARNING: to_prepare is deprecated and will be removed from Rails 5.1 (use ActiveSupport::Reloader.to_prepare instead)`

### Testing

I didn't have a great way to add a test for this, but the implementation seems pretty straightforward.  I've tested that the gem works with an app in development.  The app doesn't actually use server rendering, but it does boot and otherwise work great.

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -85,7 +85,11 @@ module React
 
         React::ServerRendering.reset_pool
         # Reload renderers in dev when files change
-        ActionDispatch::Reloader.to_prepare { React::ServerRendering.reset_pool }
+        if Gem::Version.new(::Rails::VERSION::STRING) >= Gem::Version.new("5.x")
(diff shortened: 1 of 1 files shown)


rmosolgo:
I think this behavior is covered by this test: https://github.com/reactjs/react-rails/blob/master/test/server_rendered_html_test.rb#L12

And we have a gemfile for Rails 5 (https://github.com/reactjs/react-rails/tree/master/gemfiles), so I'd say we're all clear! 

Thanks!

rmosolgo:
:ship: shipped in 1.7.1!

molawson:
Thanks!

Issue #173, opened 2015-02-14
Server side rendering (ajax call) - react_ujs not mounting component


gauravtiwari:
Hello there,

I am facing this problem from a while - I am rendering the component via an ajax call and the component is not mounting - obviously because react_ujs is missing.

Does anyone know what's the best way to solve this problem? Manually adding 'react_ujs' into partial does solve this problem but is there any other method to solve this problem - more elegantly?

``` ruby
<%= javascript_include_tag 'react_ujs', 'data-turbolinks-track' => true %>
```

Thanks

rmosolgo:
Hi @gauravtiwari , can you help me understand this problem?

When you say you're loading via ajax, you mean you're fetching content for a modal, or something like that, right?

And the page is properly rendered by the server, but then not re-mounted when it's inserted into the page?

Here's what I've done to handle this case: 
1. include `//= require react_ujs` in application.js (so that react-ujs is in the main bundle)
2. manually trigger the UJS mounting by calling `ReactRailsUJS.mountComponents()`. ( It's available since [this PR was merged](https://github.com/reactjs/react-rails/pull/117).)

Does that help in your case?

bogdan-dumitru:
Also @gauravtiwari, while @rmosolgo is right about how to fix this, I think there's also a conceptual problem with your plan. While it makes sense to have app endpoints that come fully rendered on a normal html request, to me it feels silly to rely on ajax call to the server to render perishable content using React which is designed for the browser. Why not have your ajax call just fetch data, and render your components client side at that point?

gauravtiwari:
Thanks @rmosolgo, makes sense - didn't knew about this.  @bogdan-dumitru Correct! But in this case, I am rendering the content for infinite scroll feed and I don't think if it's possible to do this. 

One another factor I think is important to consider is multiple HTTP requests, if one pre-renders the component, it saves some HTTP requests. 

Your thoughts?

bogdan-dumitru:
Don't wanna be overkill on this @gauravtiwari  (I don't get to debate these sorta things a lot, so I use any chance I get) but that's not necessarily true, depending on how you structure your components.

[...]

PR #528, opened 2016-05-10
Add a tip on how to trigger UJS mounting from js

shulyugin:
Haven't found in readme how to manually mount component in the end of ajax call. Googled answer here https://github.com/reactjs/react-rails/issues/173.


rmosolgo:
Thanks for adding it!

PR #538, opened 2016-05-25
feat(UJS) show helpful error messages for missing components

danielma:
The Problem
-----------

Currently, if you use an invalid component name in `data-react-class`,
you get a React invariant error that only tells you that the component
class was undefined, but it doesn't actually help you fix the error. I
end up needing to scroll backwards in the stack trace until I find this
part of the `react_ujs_mount.js` code and inspect the value of
`className`

screen shot 2016-05-25 at 11 36 04 am

The Solution
------------

Throw a helpful error (and nice console log message) about an invalid
className at the moment we try to mount the component. You can even right-click
and `Reveal in Elements Panel` to see exactly where the element is in the DOM!

screen shot 2016-05-25 at 11 37 23 am

### DIFF
--- a/lib/assets/javascripts/react_ujs_mount.js
+++ b/lib/assets/javascripts/react_ujs_mount.js
@@ -57,7 +57,14 @@
         var propsJson = node.getAttribute(window.ReactRailsUJS.PROPS_ATTR);
         var props = propsJson && JSON.parse(propsJson);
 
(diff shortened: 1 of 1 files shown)


rmosolgo:
🎉  thanks, that's awesome! this will prevent a lot of github issues, too!

PR #545, opened 2016-06-04
Fallback to EnvironmentContainer if assets_manifest is not supported

### LINKED ISSUE #544: Server rendering crashes in production with sprockets-rails < 2.2.2
#430 introduced the `ManifestContainer` which [uses `Rails.application.assets_manifest`](https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/manifest_container.rb#L9). This method was [introduced by `sprockets-rails` 2.2.2](https://github.com/rails/sprockets-rails/commit/3f26afbaa3d6a5ebe26403c6f42a70a1de96d1d6). If `Rails.config.assets.compile` is set to `false` (which it is by default in production) `SprocketsRenderer` [no longer uses `EnvironmentContainer`](https://github.com/reactjs/react-rails/pull/430/files#diff-1cd07a77af52dfe22d04099826f272a9R62) and causes the exception:

```
ActionView::Template::Error (undefined method `assets_manifest' for #<Xxx::Application:0x007f5464c7b180>):

  react-rails (1.6.2) lib/react/server_rendering/manifest_container.rb:9:in `initialize'
  react-rails (1.6.2) lib/react/server_rendering/sprockets_renderer.rb:70:in `new'
  react-rails (1.6.2) lib/react/server_rendering/sprockets_renderer.rb:70:in `asset_container'
  react-rails (1.6.2) lib/react/server_rendering/sprockets_renderer.rb:22:in `block in initialize'
```

I can see two possibilities:
1. Fallback to `EnvironmentContainer` for `sprockets-rails` < 2.2.2
2. Make the `sprockets-rails` dependency explicit

There are Rails version though ([e.g. 4.0.5](https://github.com/rails/rails/blob/v4.0.5/rails.gemspec)) that depend on `sprocket-rails ~> 2.0.0`.

tf:
Support for `Rails.application.assets_manifest` which is used by
`ServerRendering::ManifestContainer` has only been added in
sprockets-rails 2.2.2. Rails 4.0.5 depends on sprockets ~> 2.0.0.

* Feature detect if `assets_manifest` is present
* Add appraisal to test with rails-sprockets < 2.2.2

fixes #544


test/react/server_rendering/manifest_container_test.rb
```diff
@@ -1,8 +1,6 @@
 require "test_helper"
 
-# Rails 3.x doesn't have `application.assets_manifest=`
-# I don't think we have to support Rails 3 + Sprockets 3 anyways!
-if Rails::VERSION::MAJOR > 3
+if React::ServerRendering::ManifestContainer.compatible?
```
tf: I am leaning towards changing this to an explicit check of `Sprockets::Rails::VERSION`. While this variant is less repetitive, I am afraid some unforeseen future change of the `compatible?` method might cause the tests to be skipped for unexpected cases.

test/react/server_rendering/yaml_manifest_container_test.rb
```diff
@@ -1,6 +1,6 @@
 require "test_helper"
 
-if Rails::VERSION::MAJOR == 3
+if React::ServerRendering::YamlManifestContainer.compatible?
```
tf: Same here. Considering to revert this change.

test/react/server_rendering/manifest_container_test.rb
rmosolgo: Yes, I agree -- better to keep the tests simple so we'll see if a future change causes them to break! 

Alternatively, you could add unit tests for `.compatible?`

tf:
I've updated the PR. Checking `sprockets-rails` versions in the tests was a bit of a pain since some older versions don't even define `Sprockets::Rails::VERSION`. I still think it's an improvement, though.

rmosolgo:
Awesome, thanks so much for the patch on this bug!

PR #547, opened 2016-06-06
refactoring to reduce complexity

jethroo:
mainly moving the type lookup Hash implementation in its own module. This should also help to visualize the actual test coverage.

### DIFF
--- a/lib/generators/react/component_generator.rb
+++ b/lib/generators/react/component_generator.rb
@@ -1,5 +1,51 @@
 module React
   module Generators
+    module ComponentTypes
+      BASE = 'React.PropTypes'.freeze
+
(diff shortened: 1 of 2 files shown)


rmosolgo:
Sorry I didn't respond sooner -- is this a matter of preference, or is there a gain for end users or gem maintainers from this change?

jethroo:
hi, it is solely covering code quality. I started driven by the code climate score for this class. So i tried to refactor it to be more ruby-ish, so it should help mainly the maintainers if they want to change something here.  That is why I also prefixed it chore, its not changing any functionality (or should not) but improving how the code is organized and can be read.

rmosolgo:
This code is working well and I don't see a need for cosmetic changes. However, thanks for your offer!

Issue #550, opened 2016-06-14
View Helper Implementation is set before the initializer gets called 


bogdan-dumitru:
- [x] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.

-------

I'm trying to use a custom view helper implementation class that I've defined and set in my initializer file like so:
```ruby
  config.react.view_helper_implementation = CustomViewHelper
```
But the default ``ComponentMount`` ends up being used.
I added some breakpoints in the init process to try to figure things out and as far as I can tell the [initializer block for "react_rails.setup_view_helpers"](https://github.com/reactjs/react-rails/blob/master/lib/react/rails/railtie.rb#L28-L45) gets executed _before_ my initializer sets the config class. I added breakpoints and that's the order I saw, and it explains the behaviour, but I'm unsure if dev environment affects this somehow.

Either I'm missing something or these init blocks should run after the app gets initialised. The code there depends on other config variables as well (that end up being the defaults) and there are other blocks which would fall under the same issue. Is it a general problem or am I missing something?

rmosolgo:
I'm _guessing_ it's a general problem, but I haven't noticed it because I only use the default ones! 

Maybe that initialization should be moved to the `after_initialize` block at the bottom of the file? 

(By the way, for a workaround, could you assign the implementation directly? 

```ruby 
 React::Rails::ViewHelper.helper_implementation_class = CustomViewHelper
```

?)

bogdan-dumitru:
Yeah that's what I ended up doing, assigning the implementation directly, and it works.

bogdan-dumitru:
@rmosolgo Actually I think cleanest solution would be to add an ``:after => :load_config_initializers`` to all the initializer blocks defined there, to ensure they are called after we've set the configs in the initializer.
I'm happy to give that a go and open a pull request if it sounds good.

rmosolgo:
Yes, that would be great! I didn't know about `:load_config_initializers`, good to know!

danott:
@rmosolgo I believe this can be closed now that #551 has been merged?

rmosolgo:
yes, thanks!

PR #551, opened 2016-06-14
Run initializer blocks after :load_config_initializers

bogdan-dumitru:
As discussed in #550

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -25,7 +25,7 @@ module React
       end
 
       # Include the react-rails view helper lazily
-      initializer "react_rails.setup_view_helpers", group: :all do |app|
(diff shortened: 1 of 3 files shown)


rmosolgo:
I can't think of a way to test this, but can you confirm that it solves the problem you described in #550  ?

bogdan-dumitru:
Yeah I also was a bit disappointed in me for adding this without a spec. I can confirm that it does indeed solve it for me. I tested it from my fork without setting the helper class manually and it did set it correctly. 

But for the sake of it, let me try something. I wanna try to write a failing spec for it, open a separate PR for the failing spec just to see it fail in CI and then add the spec to this PR as well. Sounds good?

rmosolgo:
🙌 You rock, thanks for the patch!!

bogdan-dumitru:
Much obliged

PR #552, opened 2016-06-15
Add failing spec for adding view helper implementation

bogdan-dumitru:
This is just the failing spec for #551 in order to see that it fails in CI and that it doesn't in the branch with the fix.


bogdan-dumitru:
Small note: the failing message is misleading because I added the assertion arguments in the wrong order 😸

rmosolgo:
merged in #551

PR #560, opened 2016-06-29
Support .js.jsx.coffee files with Sprockets 4

djudd:


rmosolgo:
Nice, I guess this test covers `.jsx.coffee`, but it was previously failing on Sprockets 4? https://github.com/reactjs/react-rails/blob/master/test/react/jsx_test.rb#L46

I see Sprockets 4 passes now :tada: where it used to fail ... do you want to move it from the "allow failures" group in `.travis.yml`? You got it passing, so I'll maintain it :P

djudd:
Sure, I'll update travis.yml

djudd:
Thanks for the quick response! I've been upgrading the https://www.academia.edu Rails app to Sprockets 4, and among the various things that needed patching, `react-rails` wins in response time & also quality of preexisting tests. ;)

rmosolgo:
Glad to hear it, thanks for the patch!

rmosolgo:
Shipped in 1.8.0!

PR #572, opened 2016-07-27
Silence sprockets 3.7.0 deprecation warnings

chocoby:
react-rails 1.8.0 throws deprecation warnings when using sprockets 3.7.0.

```
% bin/rails c
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then use `register_compressor` or `register_transformer`.
https://github.com/rails/sprockets/blob/master/guides/extending_sprockets.md#supporting-all-versions-of-sprockets-in-processors
 (called from block in <class:Railtie> at /path/to/react-rails-1.8.0/lib/react/rails/railtie.rb:110)
Loading development environment (Rails 4.2.7)
[1] pry(main)>
```

Same problem has been fixed in rails/sass-rails#381


seuros:
deprecations should be removed or replaced,  not silenced.

chocoby:
OK, thanks!

rmosolgo:
Thanks for pointing out this issue! I'll be cutting a new version as soon as #576 is in good shape.

PR #574, opened 2016-07-28
Use new syntax for sprocket 3.7.0+

seuros:
This will remove the deprecation warning in sprockets 3.7.0+.


rmosolgo:
Awesome, thanks for taking care of this!

PR #576, opened 2016-07-29
refactor(js) stop building broken addon files

rmosolgo:
These were added to support pick-and-choose addon inclusion, but we forgot that since they include `require("react")` (etc), you'd end up with double-copies of a lot of modules, not good! So they've been undocumented for a long time, and they've never worked :S


rmosolgo:
As far as I can tell, the changes to the react.js files are due to webpack require numbers changing. I guess the numbers are shared between built files, so removing some of the build targets caused the numbers to change?

PR #582, opened 2016-08-08
add preprocessor to handle comment directives in sprocket 3.7.0+ , fixes #580

### LINKED ISSUE #580: v.1.8.1 assets precompile //=require will not be transformed

- [X] I have another issue to discuss.

-------

Hi. We have troubles with the newest release v1.8.1
In our .jsx files we do `//= require ./xyz` which gets transformed into the code from the required file in development, but not when I ran `rake assets:precompile`.

We are using `sprockets 3.7.0` and I am quite sure that the wrong behaviour is a result of following change:
https://github.com/reactjs/react-rails/compare/v1.8.0...v1.8.1#diff-ffad3c61836f28360e5c21dce8362d9fL104 (https://github.com/reactjs/react-rails/commit/99eb4e155458e6a9dc5dc08f6db2b68c1faa611e)

I've locked react-rails to v1.8.0 which resolves the problem, but I don't want to stack to this release.

seuros:
can you please check this PR @forelabs ?

### DIFF
--- a/test/react/jsx/jsx_prepocessor_test.rb
+++ b/test/react/jsx/jsx_prepocessor_test.rb
@@ -1,0 +1,16 @@
+require 'test_helper'
+
+when_sprockets_available do
+  class JSXPreprocessorTest < ActionDispatch::IntegrationTest
+    EXPECTED_JS = <<javascript
+React.createElement(\"div\", { className: \"le-javascript-child\" });
+React.createElement(\"div\", { className: \"le-javascript\" });
+javascript
+
+    test 'executes //= require directives' do
+      get '/assets/require_test/jsx_preprocessor_test.js'
+      assert_response :success
+      assert_compiled_javascript_matches(@response.body, EXPECTED_JS)
+    end
+  end
+end
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -103,9 +103,11 @@ module React
         if !sprockets_env.nil?
           if Gem::Version.new(Sprockets::VERSION) >= Gem::Version.new("3.7.0")
             sprockets_env.register_mime_type("application/jsx", extensions: [".jsx", ".js.jsx", ".es.jsx", ".es6.jsx"])
-            sprockets_env.register_transformer("application/jsx", "application/javascript", React::JSX::Processor)
             sprockets_env.register_mime_type("application/jsx+coffee", extensions: [".jsx.coffee", ".js.jsx.coffee"])
+            sprockets_env.register_transformer("application/jsx", "application/javascript", React::JSX::Processor)
             sprockets_env.register_transformer("application/jsx+coffee", "application/jsx", Sprockets::CoffeeScriptProcessor)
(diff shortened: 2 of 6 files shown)


forelabs:
[X] no comment in compiled asset
[X] comment converted into required component code

works like a charm!

rmosolgo:
@seuros what do you think about adding a test case for this? 

I'm trying to see if I understand ... since we're defining a new mime type in `react-rails`, we also have to tell Sprockets  to preprocess `//= require`s in this new mime type, right? It seems like we should add a test case to cover `//= require`s in the new mime type. 

Maybe you could add a new directory, `assets/javascripts/require_test/`, and make a few files, then add a `.jsx` file which requires them, eg 

```js 
// app/assets/javascripts/require_test/jsx_require_test.jsx
//= require ./jsx_require_child_1
//= require ./jsx_require_child_2

<div className="own-div" />
```

Then, fetch the "compiled" asset and make sure it includes all required content:

```ruby 
jsx_require_asset = Rails.application.assets["require_test/jsx_require_test"].to_s 
# Make sure the asset has required content & own content:
assert_includes jsx_require_asset, jsx_require_child_1_content
assert_includes jsx_require_asset, jsx_require_child_w_content
assert_includes jsx_require_asset, own_content 
```

Or, another way to make sure that this works the way we expect in the future?

seuros:
@rmosolgo Sure thing.

test/react/jsx/jsx_prepocessor_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+when_sprockets_available do
+  class JSXPreprocessorTest < ActionDispatch::IntegrationTest
+    EXPECTED_JS = <<javascript
+React.createElement(\"div\", { className: \"le-javascript-child\" });
+React.createElement(\"div\", { className: \"le-javascript\" });
+javascript
+
+    test 'accepts harmony: true option' do
```
rmosolgo: should be `test "executes //= require directives" do` ?

test/react/jsx/jsx_prepocessor_test.rb
```diff
@@ -0,0 +1,16 @@
...
+      get '/assets/require_test/jsx_preprocessor_test.js'
+      assert_response :success
+      assert_match(EXPECTED_JS, @response.body)
```
rmosolgo: seems like this is failing on whitespace :weary: 

maybe you could move this custom assertion to `test_helper` and use it here? https://github.com/reactjs/react-rails/blob/master/test/react/jsx_test.rb#L71-L77

seuros: Oh, i didn't see it. doing that now.

seuros:
@rmosolgo we can remove the work in progress in sprocket 4 now.

rmosolgo:
🎉  thanks, i'll merge and ship 1.8.2 pronto.

rmosolgo:
removed the sprockets 4 note in 5049473

rmosolgo:
:ship: 1.8.2  🎊  thanks again!

PR #595, opened 2016-08-30
Allow configurability of React/ReactDOM instance per component

j-clark:
Hi there,

This isn't a complete working/tested implementation. I'm just testing the waters to see if you guys would be amenable to a PR looking something like this.

**An example problem**

Imagine I've got 2 different widgets libraries being managed by 2 different teams. `widgets-core` is managed by 1 team, and is full of tried and true widgets built with React. `widgets-experimental` is full of the latest widgets being developed by a different team, also with React. Both teams deliver a .js file globally exposing their widgets and are separate from the Rails team.

I'm on the Rails team and everything is going great in my Rails app using `react-rails` until I start to use widgets form both `widgets-core` and `widgets-experimental` on the same page. They use different versions of React. The components are already unhappy with being rendered by a different version of React than they were defined with, but to make it worse, `widgets-core` is still pre-0.14 and `widgets-experimental` is using React 15.x, meaning that these versions of React have different rendering interfaces! I'm stuck!
## 

Hopefully that example wasn't too contrived. It's very similar to a problem that I'm experiencing. This PR represents a potential way to solve the problem. Imagine a Rails partial that looks like this

``` slim
= react_component('Widgets.Core.PrimaryWidget', 
  props, 
  { react_name: 'Widgets.Core.React',
    react_dom_name: 'Widgets.Core.React' })

= react_component('Widgets.Experimental.WizzBang', 
  props, 
  { react_name: 'Widgets.Experimental.React',
    react_dom_name: 'Widgets.Experimental.ReactDOM' })
```

This allows us our React components to be developed independently from the Rails app and to package up the React versions they need in order to render correctly.

Thoughts?

Totally open to suggestions and, of course, to renaming / shifting code around / whatever to conform to codebase style guides.

### DIFF
--- a/lib/assets/javascripts/react_ujs_mount.js
+++ b/lib/assets/javascripts/react_ujs_mount.js
@@ -11,6 +11,9 @@
     // example: `data-react-props="{\"item\": { \"id\": 1, \"name\": \"My Item\"} }"`
     PROPS_ATTR: 'data-react-props',
 
+    REACT_ATTR: 'data-react-name',
+    REACT_DOM_ATTR: 'data-react-dom-name',
+
     // helper method for the mount and unmount methods to find the
     // `data-react-class` DOM elements
     findDOMNodes: function(searchSelector) {
@@ -77,6 +80,8 @@
         var constructor = this.getConstructor(className);
         var propsJson = node.getAttribute(window.ReactRailsUJS.PROPS_ATTR);
         var props = propsJson && JSON.parse(propsJson);
+        var reactDomInstance = getReactDomInstanceAtNode(node);
+        var reactInstance = getReactInstanceAtNode(node);
 
         if (typeof(constructor) === "undefined") {
           var message = "Cannot find component: '" + className + "'"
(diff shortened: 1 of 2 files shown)


rmosolgo:
Wow, that's a tight spot! 

I'd like to support your case, but in a way that minimizes maintenance overhead for the library. 

In #503, one little section of `mountComponents` was broken out into a separate function, `getConstructor`. This way, people could override that function if they wanted to lookup constructors in a new way. Could we apply a similar approach here? 

For example, perhaps this line could be extracted to a separate function: 

- from 

  ```js
  ReactDOM.render(React.createElement(constructor, props), node);
  ```

- to 
  
  ```js 
  renderComponent: function(node, constructor, props) { 
    ReactDOM.render(React.createElement(constructor, props), node);
  },
  ```

Then, in your case, you could provide a custom `renderComponent` function: 

```js 
ReactRailsUJS.renderComponent = function(node, constructor, props) { 
  if (node.getAttribute("data-react-experimental")) {
    Widgets.Experimental.React // do whatever 
  } else {
    Widgets.Core.React // do whatever 
  }
}
```

And I think the view helper will pass any "unknown" properties on to the DOM element so 

```erb 
<%= react_component("MyComponent", {}, {data: {react_experimental: true}}) %>
<!-- I think this will become: -->
<div ... data-react-experimental="true" />
```

Would an extension point like that  work in this case?

rmosolgo:
I don't think this approach is a good fit for the gem in general, but if you get a chance to extract those UJS functions, please reopen!

PR #597, opened 2016-09-01
Fix alt asset path

wadewinningham:
Allow for a non-default asset path for compiled assets.

On initial deployments to new servers, skip files we haven't had the opportunity to compile yet.

### DIFF
--- a/lib/react/server_rendering/yaml_manifest_container.rb
+++ b/lib/react/server_rendering/yaml_manifest_container.rb
@@ -6,12 +6,12 @@ module React
     # but sometimes, they're compiled to other directories (or other servers)
     class YamlManifestContainer
       def initialize
-        @assets = YAML.load_file(::Rails.root.join("public/assets/manifest.yml"))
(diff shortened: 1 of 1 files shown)


rmosolgo:
That change to use the configured prefix looks great, thanks! 

Can you help me understand the change which returns `""` when an asset is not found? Why is that desirable? 

In my own case, if an asset is not available, I'd rather have an error than a silent failure. But I don't use a YAML manifest -- is it different in this case?

wadewinningham:
Sorry, I accidentally submitted this PR here rather than my fork of it.

As for returning `""`, that's a work in progress. The issue we're running into is during deployment. In our Rails 3.2 app, deploying to a new server with the `react-rails` gem gets to the `asset:precompile` step and fails because it's looking for the `server_rendering.js` file in `manifest.yml`. The problem is that the particular check is prior to the asset being compiled in the first place.

RIght now, by returning `""` (which is likely not the most elegant solution) we can successfully deploy and afterwards `react-rails` seems to work as intended. I'm working on this though to see if there's a better solution and will submit a proper PR once we get there.

rmosolgo:
Oh, I see, good luck, I'll keep an eye out for any updates!

PR #603, opened 2016-09-27
Add ability to set prerender to false when calling from controller

jamesmccann:
This change allows the `render` call in the controller to pass a `prerender` flag which overrides the default of true for a controller render.

If the React component is the only DOM rendered for a view it makes sense to render it from the controller rather than have a one-line view ERB file that calls `react_component`, and it should be possible to do this without having to force that component to be server rendered.

### DIFF
--- a/lib/react/rails/controller_renderer.rb
+++ b/lib/react/rails/controller_renderer.rb
@@ -3,7 +3,8 @@ module React
     # A renderer class suitable for `ActionController::Renderers`.
     # It is associated to `:component` in the Railtie.
     #
-    # It is prerendered with {React::ServerRendering}.
+    # It is prerendered by default with {React::ServerRendering}.
+    # Set options[:prerender] false to disable prerendering.
     #
     # @example Rendering a component from a controller
     #   class TodosController < ApplicationController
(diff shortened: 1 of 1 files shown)


rmosolgo:
Nice! It makes sense to provide an option here instead of forcing people to use `prerender: true`. To make sure I understand correctly, this _allows_ you to 

```ruby
render component: "MyPage", prerender: false
```

And in that case, the response would be equivalent to:

```erb
<%= react_component("MyPage") %>
```

(Just a `<div data-react-class="MyPage" data-react-props="{}"></div>` ?)

Have I understood correctly? 

In any case, this feature would be nicely documented and protected against regression by adding a test case, do you mind doing that? 

You might be able to parameterize the [current example of `render component:`](https://github.com/reactjs/react-rails/blob/master/test/dummy/app/controllers/server_controller.rb#L18-L25) to accept `params[:prerender]`.  Then, you could copy & modify the [test for that controller action](https://github.com/reactjs/react-rails/blob/master/test/server_rendered_html_test.rb#L56-L69), calling it with `prerender=false` and asserting that the resulting HTML contains what you expect it contains. 

That's how I'd try it anyways, but if you see another way, please help yourself!

jamesmccann:
@rmosolgo that's right, yes :)

Thanks for those pointers on doc/tests I'll check that out.

perezperret:
Is there a reason why this still hasn't been merged? @jamesmccann

rmosolgo:
It still needs a test

xinranxiao:
@jamesmccann are you planning to write a test for this?

rmosolgo:
This feature was just merged, I'll release 1.11.0 with it tomorrow, thanks for your work!

PR #604, opened 2016-09-27
Add code option to Sprockets Render

NGMarmaduke:
This enables additional code to be added to the Exec JS context when initialised.

Our app is consuming assets generated via Sprockets and Webpack, this option allows the Webpacked assets to injected along side the Sprockets assets for use in the ExecJS runtime.

### DIFF
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -17,6 +17,7 @@ module React
         @replay_console = options.fetch(:replay_console, true)
         filenames = options.fetch(:files, ["react-server.js", "components.js"])
         js_code = CONSOLE_POLYFILL.dup
+        js_code << options.fetch(:code, '')
 
         filenames.each do |filename|
           js_code << asset_container.find_asset(filename)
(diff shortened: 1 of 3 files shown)


rmosolgo:
Very nice! For user code, does this option go in `server_render_options`, for example: 

```ruby 
  code: File.read(webpack_bundle),
  # ...
}
```

?

NGMarmaduke:
It could certainly be set via the server_render_options in `application.rb`, for our app we already use a custom render to enable i18n:

```ruby
class JsRenderer < SprocketsRenderer
  def initialize(options = {})
    options[:code] = webpack_assets
    super(options)
  end

  def before_render(component_name, props, prerender_options)
    super + set_i18n_js
  end

  def after_render(component_name, props, prerender_options)
    super + 'alt.recycle();'
  end

  private

  def set_i18n_js
    [i81n_js, moment_js].join
  end

  def i81n_js
    "I18n.locale = '#{ I18n.locale }';"
  end

  def moment_js
    "moment.locale(I18n.locale);"
  end

  def webpack_assets
    webpack_manifest.javascript_assets.map do |filename|
      webpack_manifest.find_asset(filename)
    end.join
  end

  def webpack_manifest
    WebpackAssets.manifest
  end
end
```

rmosolgo:
Very cool, thanks for sharing your use case!!

rmosolgo:
Just shipped in 1.9.0!

Issue #614, opened 2016-10-14
ExecJS environment defines window


NGMarmaduke:
- [x] My app crashes with `react-rails`, so I've included the stack trace and the exact steps which make it crash.

---

The `ExecJSRenderer` defines window as the global scope for server rendering here: [ExecJSRenderer:36](https://github.com/reactjs/react-rails/blob/8bea8e1bbdec3e56485c054aabfd0d4a1029be47/lib/react/server_rendering/exec_js_renderer.rb#L36)

This can cause major issues when using 3rd party libraries that perform type of checks on the window variable to detect the presence of a browser environment.

I have created an example Rails app to reproduce the issue:
https://github.com/NGMarmaduke/react-rails-window-example

The sample code with the window check is from yahoo's [subscribe-ui-event](https://github.com/yahoo/subscribe-ui-event/blob/2fe764d1b5f214af5125b3553209f0bac3d52ca4/src/mainEventConnectors.js#L30-L34)

The resulting error is:

```
ExecJS::ProgramError
  ReferenceError: document is not defined
```

Realistically window shouldn't be defined or available on the server, initial investigation of why this is being defined points me at [react-builds/react-server.js](https://github.com/reactjs/react-rails/blob/master/react-builds/react-server.js), it is assigning variables on the window:

``` javascript
window.React = React;
window.ReactDOM = ReactDOM;
window.ReactDOMServer = ReactDOMServer;
```

As an aside `ReactDOM` isn't required for server rendering.

PR #611, opened 2016-10-04
feat(react.js) update to 15.3.2

rmosolgo:
Updated `package.json` and ran `rake react:update`

PR #615, opened 2016-10-14
Remove window variable from ExecJS environment

NGMarmaduke:
Fixes https://github.com/reactjs/react-rails/issues/614

Defining window on the server can cause major issues when using 3rd party libraries that perform type of checks on the window variable to detect the presence of a browser environment.

A common check for presence of the browser is:

``` javascript
if (typeof window !== 'undefined') { 
  // do browser stuff
}
```

This PR removes the requirement for window to exist on the server. It will however be a breaking change for applications that implicitly expect the existence of window in all javascript environments. 

The definition of window on the server could potentially be made a configuration option to mitigate errors.

~~This PR also updates React to version 15.3.2~~


NGMarmaduke:
Looks like running into the same error as https://github.com/reactjs/react-rails/pull/611

NGMarmaduke:
Removed the update to React

rmosolgo:
I think this is Ok, but we need to document the change so that people who currently _depend_ on window (using it as the global namespace) know how to upgrade. 

I checked coffeescript, it looks like it compiles to something compatible: 

```coffee
@test = false 
```

to 

```js
(function() { 
  this.test = false 
}).call(this)
```

(I thought it might use `window` as `@`, but looks like not)

NGMarmaduke:
Indeed, I'm attempting to apply this update to our app to see if its even feasible / sensible to make this change

rmosolgo:
How did it go applying this to your app?

rmosolgo:
I hope it went ok :P But I agree, making server render _pretend_ to be a browser env only makes it worse down the line.

rmosolgo:
🚢  in `1.10.0`, thanks!

dbackeus:
Not actually sure how to fix the errors we get after upgrading to `1.10.0`:
```
     Failure/Error: <%= react_component "ComponentName", {
     
     ActionView::Template::Error:
       ReferenceError: window is not defined
     # (execjs):44062:3
     # (execjs):44064:2
```

The components we use do not reference `window` explicitly as far as I can see (but are defined as `@ComponentName = React.createClass(...)`).

The CHANGELOG mentions manually adding `window = this` but not sure where to do that?

rmosolgo:
@dbackeus Do those errors come from server rendering? 

Do you have a server-rendering manifest (some file that requires all the components for server rendering)? If so, you can add it to that file, for example: 

```js 
//= require components 
// polyfill `window` variable for server rendering environments:
var window = this 
```

If that doesn't help, can you share the lines of code that reference `window`? A better (or earlier) polyfill may be needed

piratefsh:
FYI I had the same `window is not defined` issue and tried to set `var window = this` after the components were required but that still gave the error. What worked was to set `window` before the React was required. In my case, React was required in my `react_app` file.

e.g.

```
// polyfill for prerendering. must be set before requiring react
var window = this;

//= require react_app
```

piratefsh:
Since `window` is needed to even load React, shouldn't this be mentioned in the README setup?

PR #618, opened 2016-10-22
polyfill clearTimeout and setTimeout for server side rendering

krzysiek1507:


krzysiek1507:
@rmosolgo :)

rmosolgo:
Sorry for the delay, I was trying to think if this would mess with anyone who has already polyfilled them, or if it would affect new users who don't know it's polyfilled. But I think a user polyfil would override this one, and this error is better than "undefined is not a function"!

krzysiek1507:
@rmosolgo thanks!

rmosolgo:
🚢  in `1.10.0`, thanks!

PR #619, opened 2016-10-22
Use ['default'] property accessor, not .default

Igor10k:
Using reserved words in dot property accessors is allowed in ES5,
but not in ES3. Android 2 throws `SyntaxError: Parse error` with `.default`


rmosolgo:
Thanks! I'm not sure how to add a regression test for this issue, but I'll try to keep this requirement in mind for the future.

rmosolgo:
🚢  in `1.10.0`, thanks!

PR #621, opened 2016-10-24
Reactrb is now part of Ruby Hyperloop

Willianvdv:
> Reactrb is now a part of Ruby Hyperloop

http://ruby-hyperloop.io/reactrb/


rmosolgo:
Cool, thanks for updating that!

Issue #626, opened 2016-10-28
Allow Subscribe to Redux Store in react_component helper


BreahnaDavid:
Help us help you! Please choose one:
- [ ] My app crashes with `react-rails`, so I've included the stack trace and the exact steps which make it crash.
- [ ] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.
- [x] I'm trying to use `react-rails` with another library, but I'm having trouble. I've described my JavaScript management setup (eg, Sprockets, Webpack...), how I'm trying to use this other library, and why it's not working.
- [x] I have another issue to discuss.

---

We want to use redux. http://redux.js.org/
But we need to subscribe each our component, so it looks that we have to create our own helper method for it, and perhaps to have the same code as in react_ujs_mount.js
why not allow us and another people who use Redux to subscribe on react_component?

rmosolgo:
I made a suggestion on your PR -- perhaps we could accept a user-defined after-render hook, would that help in this case? 

Another approach is to pass the store name as a prop, eg 

```erb
<%= react_component("ItemList", {storeName: "Items"}) %>
```

Then subscribe in the component: 

```js
componentDidMount: function() {
  this._store = App.Stores[this.props.storeName]
  this._store && this._store.subscribe(this.onStoreChange)
} 
componentWillUnmount: function() {
  this._store && this._store.unsubscribe(this.onStoreChange)
}
```

alkesh26:
As per the suggestion by @rmosolgo, we can close this issue.

PR #627, opened 2016-10-28
Allow Subscribe to Redux Store in react_component helper

BreahnaDavid:
https://github.com/reactjs/react-rails/issues/626


rmosolgo:
This is an interesting use case! But I don't think we should include state management in this gem. 

However, if you find a general improvement (not Redux-specific, and not requiring the user to set global values on `window`), feel free to suggest it in a separate PR. For example, I see you're adding a callback to `ReactDOM.render`.  Is there a way we could support that in a general way? For example, accepting a user-defined function as `ReactRailsUJS.afterRender` and calling that function if it's present?

Issue #639, opened 2016-11-22
Indirection inside server side renderers


prikha:
Help us help you! Please choose one:

- [ *] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.
-------

Hi guys we are running react-rails in production for a while. We have some global JS context, and we do implement a custom renderer. 

However there are strange things inside ServerRendering::SprocketsRenderer

look over here:

https://github.com/reactjs/react-rails/blob/master/lib/react/server_rendering/sprockets_renderer.rb#L43

```
def render(component_name, props, prerender_options)
  ...
  super(component_name, props, {render_function: react_render_method})
end
```

Imagine I would implement some custom prerender (and I really do so) options - and looking and the renderer itself - i would expect the same `prerender_options` entered `render` method - to be passed to `before` and `after` hooks.

The code above shows that it is not true, we are cutting those options and passing something `{render_function: react_render_method}`.  I think it should be changed. And if it sounds reasonable - I can submit a PR

rmosolgo:
:+1: Your proposed change sounds good to me! That will make custom renderers much more flexible.

prikha:
Ok, I`ll prepare PR

prikha:
Hi @rmosolgo please review https://github.com/reactjs/react-rails/pull/641

prikha:
Updated PR as required
@rmosolgo FYI

rmosolgo:
Thank you for this improvement, I'll cut a new version soon!

PR #641, opened 2016-11-28
#639 Pass pre-render options to callback functions

prikha:
Changes:

- Refactor a bit base renderer
- Refactor `SprocketsRenderer` to preserve custom prerender options

#639

### DIFF
--- a/lib/react/server_rendering/exec_js_renderer.rb
+++ b/lib/react/server_rendering/exec_js_renderer.rb
@@ -11,16 +11,10 @@ module React
       end
 
       def render(component_name, props, prerender_options)
-        render_function = prerender_options.fetch(:render_function, "renderToString")
-        js_code = <<-JS
-          (function () {
-            #{before_render(component_name, props, prerender_options)}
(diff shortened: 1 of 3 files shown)


rmosolgo:
This patch looks great, but can you add a test to demonstrate the changed behavior?

Sorry about the CI crash. I've updated `.travis.yml`, do you mind rebasing on latest master?

prikha:
Yepp, I`ll try something out

prikha:
Done that(https://travis-ci.org/reactjs/react-rails/builds/180025155), however the build on Travis is extremely slow. Why would you keep 

```
allow_failures:
    - rvm: jruby-9.0.1.0
```

which always take 10 minutes and always fail?

rmosolgo:
I agree that the JRuby situation is not good! I think we moved them into "allow failures" for Rails 5, since no JRuby version supported Rails 5 at that time. 

It looks like they timeout now :( It would be nice to fix those, because I think we _should_ support JRuby.

Thanks for this improvement! It will make custom renderers more powerful!

rmosolgo:
🚢  in `1.10.0`, thanks!

PR #642, opened 2016-11-28
Allow `camelize_props` to be passed into component mounter

NGMarmaduke:
This allows the use of camelize props on a case by case basis; useful if you want the option without breaking existing components

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,12 @@ MyApp::Application.configure do
 end
 ```
 
+or when mounting:
+
+```erb
+<%= react_component('HelloMessage', {name: 'John'}, {camelize_props: true}) %>
+```
+
 ### Rendering components instead of views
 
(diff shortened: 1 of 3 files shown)


rmosolgo:
:+1: sorry about the CI failure, I've just updated the `travis.yml` on master. Would you mind rebasing & force pushing so we can see the sweet, beautiful ✅  ?

NGMarmaduke:
@rmosolgo thanks for the heads up, build's green again

rmosolgo:
:metal: I just realized, this is kinda weird because you _can_ override default `false` by passing `true`, but if the default is `true`, you can't override with false. I'm going to PR with that behavior in a moment!

rmosolgo:
🚢  in `1.10.0`, thanks!

Issue #640, opened 2016-11-22
content_tag method expected behaviour is broken


prikha:
Help us help you! Please choose one:

- [*] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.
-------

Hi guys, and again I would like to ask if there are anyone using view helper `react_component` like so:

```
# slim template
= react_component 'PhoneBlock', {}, {} do
  ' Arbitrary content here
end
```

Apparently this one gives unexpected results: 

Text `Arbitrary content here` would appear on the page twice. I haven't find a good solution yet, but the roots of the problem lay in the field of view context_buffer. 

We do however use this hack:
```
def react_component(name, props = {}, options = {}, &block)
    ...
    if block_given?
      predefined_content = capture(&block)
      helper_obj.react_component(name, props, options) { predefined_content }
    else
      helper_obj.react_component(name, props, options)
    end
  end
```

That gives predictable results. We capture block output inside the view context and just pass it down.

Do you have any thoughts/directions on this?

rmosolgo:
I don't use the view helper in that way, but I agree with you that predictable results are better! A patch would be welcome if you have the time for it :)

prikha:
Sure I`ll take a look this week.

prikha:
Hi @rmosolgo please review #644

rmosolgo:
Thanks for fixing this!

PR #644, opened 2016-11-30
Fix react_component helper usage with a block

prikha:
Capture block output to prevent double rendering.
references #640


rmosolgo:
Awesome, thanks!

rmosolgo:
🚢  in `1.10.0`, thanks!

henrik:
@prikha Thank you for this! Got this error when upgrading to 1.3.3 (small steps).

PR #645, opened 2016-12-01
feat(ComponentMount) allow overriding camelize_props=true with local false

rmosolgo:
A new feature in #642 was to pass `camelize_props: true` to override the default of `false`

This adds the ability to `camelize_props: false` to override the default of `true`.


NGMarmaduke:
@rmosolgo Nice, looks good 👍

PR #646, opened 2016-12-01
feat(react.js) build with yarn; update to React 15.4.1

rmosolgo:
Trying again with the latest versions


Rakefile
```diff
@@ -40,7 +40,7 @@ namespace :react do
   desc "Use NPM to install the JavaScript dependencies"
```
krzysiek1507: `Use yarn`? ;)

rmosolgo: Thanks!

PR #659, opened 2017-01-20
Correct Readme installation section

novikserg:


rmosolgo:
The install generator [creates `components.js`](https://github.com/reactjs/react-rails/blob/master/lib/generators/react/install_generator.rb#L45-L49), which that line requires.

Why should this be changed?

novikserg:
@rmosolgo hi Robert, I used Rails 5.0.1, ruby 2.2.4 and installation instructions in readme. Generator resulted in generating `components/` folder (not file), and giving me an assets error for `require components`. As you would expect, it started working after I changed to `require_tree ./components`. So not sure if it's just Rails 5 issue or not, but it definitely exists

akshaymohite:
@novikserg I just tried with rails `5.0.1` and react-rails `1.10.0`, and could not reproduce such issue.

I don't believe the issue would be related to ruby `2.2.4`.

Sample repository just for your reference: https://github.com/akshaymohite/rails-5-react-rails-test

rmosolgo:
Thanks for taking another look, I guess we'll close this unless we hear someone else runs into it!

PR #662, opened 2017-01-28
Relay support

thijsc:
We'd like to start using [Relay](https://facebook.github.io/relay/) and currently use `react-rails`. I've looked at various options to use both Relay and this gem it seems that adding a build to this gem is the only viable way if you want to stick with the asset pipeline.

This pull is a proof of concept, if you think this is a good direction I'll finish it. It already builds a separate JS file that includes Relay. The next step would be to add this variant to `AssetVariant`.

Are you open to adding support for Relay in the gem? If so, any pointers on how to continue with this pull?


thijsc:
So it turns out this is actually not that simple. We'd need to add a setup for queries similar to the current one for JSX. For the moment we've moved to https://github.com/Shopify/sprockets-commoner so we can use Babel to do the various conversions.

rmosolgo:
Thanks for giving it a shot! I imagine the transform step is a bit tricky, glad commoner was able to help!

PR #667, opened 2017-02-23
remove coffee-script-source from dependency

chulkilee:
coffee-script-source is not required to use this gem, so remove it from dependency.


rmosolgo:
Thanks for cleaning this up!

PR #672, opened 2017-03-20
Add ability to set prerender to false when rendering from controllers

xinranxiao:
More or less https://github.com/reactjs/react-rails/pull/603 with tests

@rmosolgo could take a look at this when you get a chance? Thanks!!

For the test, I just wrote a case where we pass `prerender: false` as an option and changed up the expects to account for no server-side rendering.

### DIFF
--- a/lib/react/server_rendering/sprockets_renderer.rb
+++ b/lib/react/server_rendering/sprockets_renderer.rb
@@ -53,7 +53,7 @@ module React
       # default Rails setups.
       #
       # You can provide a custom asset container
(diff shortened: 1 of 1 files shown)


rmosolgo:
Sorry 😖  I tried to push another commit to refactor the test a bit, but I think I clobbered your master branch, and now I can't reopen this.

I've grabbed your commits here and will merge them as soon as the tests finish: 

https://github.com/reactjs/react-rails/pull/680

Sorry about the confusion, and thanks for taking care of this and updating the docs! Hoping to cut a new gem version this afternoon.

PR #678, opened 2017-04-03
refactor(UJS) make it webpack-friendly

rmosolgo:
As part of supporting webpacker: 

- [x] build UJS with webpack 
- [x] test it ~~with Jest~~ (jest didn't work because it doesn't support `require.context`)
- [x] publish it on NPM  https://www.npmjs.com/package/react_ujs


rmosolgo:
These broken tests were fixed on 2.0-dev in d62f518

PR #682, opened 2017-04-04
feat(ServerRendering) reload when new files are added; support configurable reloader globs

rmosolgo:
Using `FileUpdateChecker` will catch new files too, and we can expose an API for watching _any_ files (spoiler alert, I'm going to use this for webpacker 😆 )


rmosolgo:
(I got a CI pass by rerunning it)

PR #683, opened 2017-04-05
Webpacker Server Rendering

rmosolgo:
Support server rendering based on a pack called `server_rendering.js`

- [x] Load JS from packs 
- [x] Load JS from webpack-dev-server
- [x] Watch js in packs 
- [ ] Make sure files from webpack-dev-server work with ExecJS (they dont because they expect to be eval'd by a browser)


rmosolgo:
I'll circle back for the webpack-dev-server asset issue

PR #684, opened 2017-04-05
Support webpacker in generators

rmosolgo:
- [x] Component generator outputs to the configured webpacker source
- [x] Install generator adds to webpacker files


rmosolgo:
While writing the template for the `server_rendering.js` pack, I realized the current JS call for server rendering won't work, because it's coupled to component-names-as-globals. So I rejiggered it a bit, now it calls a method on `ReactRailsUJS` instead.

brentdodell:
This was broken, I believe [this](https://github.com/rails/webpacker/pull/636) Webpacker change. `Webpacker::Configuration` should now be `Webpacker.config`

PR #696, opened 2017-04-17
fix(ReactUjs): Fixed variable typo in findDOMNodes

tannermares:
### Why this was needed
 Looks like `findDOMNodes` has a typo in the argument name [here](https://github.com/reactjs/react-rails/blob/master/react_ujs/index.js#L23). `selector` should be `searchSelector`. Since the incorrect value is used ALL components are unmounted when `findDOMNodes` is called from [unmountComponents](https://github.com/reactjs/react-rails/blob/master/react_ujs/index.js#L100)

### How I fixed it
Updated argument name from `selector` to `searchSelector` [here](https://github.com/tannermares/react-rails/blob/4f65848239fd6665d8c4a8aa5878103764d83e11/react_ujs/index.js#L23)

I adding a spec. I'm not familiar with MiniTest or server side rendering and tried to implement the failure in the simplest way possible.

I believe the failing specs are from JS warnings the `rake react:update` caused. Looks like maybe the `createClass` deprecation? Should I not `rake react:update`?

Any suggestions on that front and I'll gladly implement. :)

> Note: This is my first contribution and couldn't find a "how to contribute" section. I apologize for any missing steps or documentation.

### DIFF
--- a/lib/assets/javascripts/react_ujs.js
+++ b/lib/assets/javascripts/react_ujs.js
@@ -238,7 +238,7 @@ var ReactRailsUJS = {
 
   // helper method for the mount and unmount methods to find the
   // `data-react-class` DOM elements
-  findDOMNodes: function(selector) {
+  findDOMNodes: function(searchSelector) {
     var classNameAttr = ReactRailsUJS.CLASS_NAME_ATTR
     // we will use fully qualified paths as we do not bind the callbacks
     var selector, parent;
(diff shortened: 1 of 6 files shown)


rmosolgo:
I think you're right about `react:update` causing the issues. This library isn't ready for the latest version!

I don't want to merge the failing specs, but either of these three is ok with me: 

- Correct the typo in `react_ujs/index` but don't modify the test suite 
- Correct the typo and figure out how to make the proper test case pass 
- Correct the typo & test it, _and_ update React and make the tests pass (_shudder_)

Does one of those sound good to you?

tannermares:
@rmosolgo updated

rmosolgo:
:metal: Thanks! I'm going to try to get #697  in before a release tomorrow morning.

rmosolgo:
🚢  in 2.1.0!

PR #699, opened 2017-04-19
Removed createClass as its deprecated

AG-Systems:
`React.createClass` will be removed in version 16.

https://github.com/reactjs/react-modal/issues/367

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,13 +99,6 @@ Now, you can create React components in `.jsx` files:
 ```js
 // app/assets/javascripts/components/post.jsx
 
-window.Post = React.createClass({
-  render: function() {
-    return <h1>{this.props.title}</h1>
-  }
-})
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -395,7 +388,7 @@ rails g react:component Post title:string published:bool published_by:instanceOf
 would generate:
 
 ```js
-var Post = React.createClass({
+class Post extends React.Component{
```
rmosolgo: If we change the documentation for ComponentGenerator, should we also update the code to match the new documentation?

AG-Systems: I think its a great idea

BookOfGreg:
Thank you for your contribution!
I've merged your commits into #798 so it can all go as one cohesive patch.

PR #703, opened 2017-04-29
Camelize the file name when webpacker is defined

Najtmare:


rmosolgo:
Thanks for this fix! I'll release a new version in just a bit with your changes.

PR #708, opened 2017-05-11
check for turbolinks classic before teardown

glosie:
When using Turbolinks 5+, `ReactRailsUJS.detectEvents()` throws the following error when  removing the existing event handlers:
```
Uncaught TypeError: Cannot read property 'CHANGE' of undefined
    at Object.teardown (turbolinksClassic.js?6757:9)
    at module.exports (detect.js?5263:12)
    at Object.detectEvents (index.js?c0e8:114)
```
Since Turbolinks 5 doesn't define `Turbolinks.EVENTS`, we can check for its existence before calling the classic teardown method.

### DIFF
--- a/react_ujs/src/events/detect.js
+++ b/react_ujs/src/events/detect.js
@@ -9,7 +9,9 @@ var turbolinksClassicEvents = require("./turbolinksClassic")
 module.exports = function(ujs) {
   if (ujs.handleEvent) {
     // We're calling this a second time -- remove previous handlers
-    turbolinksClassicEvents.teardown(ujs)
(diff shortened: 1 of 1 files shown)


glosie:
The Rails 5 build failures look to be related to https://github.com/rails/rails/pull/29022

rmosolgo:
Wacky about those failures, thanks for the link! 

:+1:

rmosolgo:
:ship: 2.2.0 !

glosie:
Thanks for the quick merge and release! Loving the new webpacker support ❤️

PR #729, opened 2017-05-25
Configure watched files after rails initializers have loaded

kernow:
Rails convention is now to load application config from initializers rather than `config/application.rb`. This PR makes it possible to specify`server_renderer_directories` (amongst other settings) and have the settings honoured by the file watcher. Previously some settings configured in an initializer were ignored as the file watcher block was run before initializers were loaded.

### DIFF
--- a/test/react/rails/railtie_test.rb
+++ b/test/react/rails/railtie_test.rb
@@ -1,0 +1,10 @@
+require "test_helper"
+
+class RailtieTest < ActionDispatch::IntegrationTest
+  test "reloaders are configured after initializers are loaded" do
+    @test_file = File.expand_path("../../dummy/app/pants/yfronts.js", File.dirname(__FILE__))
+    FileUtils.touch @test_file
+    results = Dummy::Application.reloaders.map(&:updated?)
+    assert_includes(results, true)
+  end
+end
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -24,7 +24,7 @@ module React
       config.react.view_helper_implementation = nil # Defaults to ComponentMount
 
       # Watch .jsx files for changes in dev, so we can reload the JS VMs with the new JS code.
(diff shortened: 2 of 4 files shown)


rmosolgo:
👏  wow, thanks this looks great! let's see how CI likes it 😬 

Nice to know I can have react-rails watch my `pants` directory 😆

kernow:
Humm, lots of failures :( it's really hard to test this change as most of the methods for the active support file watchers are private and it's not possible to get any information about which files have been added to the watch list without accessing private data which is not consistent across all of the available file watchers.

The only other option I can think of to test this is to create a mock file watcher object for the test.

test/react/rails/railtie_test.rb
```diff
@@ -0,0 +1,14 @@
+require "test_helper"
+
+class RailtieTest < ActionDispatch::IntegrationTest
+  test "reloaders are configured after initializers are loaded" do
+    begin
+      @test_file = File.expand_path("../../dummy/app/pants", File.dirname(__FILE__))
```
rmosolgo: This file name, does it end in `.js` or `.jsx`? I think only those files are watched due to `config.react.server_renderer_extensions`. Could that be why the test doesn't pass?

kernow: Good point, I'd forgotten about the file extension. Although some of the tests are passing and they passed locally which is odd as you would expect it not to pass at all. I'll check it out

kernow:
@rmosolgo this is much happier now. I assumed that all the builds would be passing on travis (before my changes) but it looks like there are several test that have not been passing on master. Is there anything else you want me to do on the PR?

rmosolgo:
No, I have to figure those ones out separately! I'll cut a release soon, I'm hoping to merge one more fix...

rmosolgo:
Finally shipped in 2.2.1!

PR #761, opened 2017-08-04
Update camelize props anchor link in README

robwilliams:
The heading 'Camelize Props` has the id `camelize-props`, not `camelize_props`


rmosolgo:
Thanks for taking care of this!

PR #768, opened 2017-08-11
CoffeeScript React component generator is fixed

ViXP:
CoffeeScript React component generator is fixed to modules export of component and to require the React object.
`rails g react:component Sample --coffee`
Before:
```
class @Sample extends React.Component
  render: ->
    `<div />`
```
After:
```
React = require('react')

class @Sample extends React.Component
  render: ->
    `<div />`

module.exports = @Sample
```


lib/generators/react/component_generator.rb
```diff
@@ -125,15 +125,23 @@ def component_name
 
       def file_header
         if webpacker?
-          %|var React = require("react")\n|
+          if options[:coffee]
+            %|React = require('react')\n|
+          else
+            %|var React = require("react")\n|
```
BookOfGreg: Would these work as import rather than requires if it's webpack?

lib/generators/react/component_generator.rb
```diff
@@ -125,15 +125,23 @@ def component_name
...
         end
       end
 
       def file_footer
         if webpacker?
-          %|module.exports = #{component_name}|
+          if options[:coffee]            
+            %|module.exports = @#{component_name}|
+          else
+            %|module.exports = #{component_name}|
```
BookOfGreg: Similar to above, could these be export default rather than module.exports?

ViXP:
Yeah, I've corrected the code you mentioned, it is right to stick to one of the possible syntaxes. Also, PropTypes was removed from base 'react' source code, so now it must be imported in such way: `import PropTypes from 'prop-types'`. I've added this line to the generator too.

BookOfGreg:
I've merged your code and commits into #798 and corrected the associated tests. Thank you for your help with Coffeescript! Could you test out that branch and see if it meets your needs?

PR #777, opened 2017-08-31
Support singleton webpacker (3 and above)

### LINKED ISSUE #775: rails generate react:install not working with Webpacker

- [x] I'm trying to use `react-rails` with another library, but I'm having trouble. I've described my JavaScript management setup (eg, Sprockets, Webpack...), how I'm trying to use this other library, and why it's not working.
- [ ] I have another issue to discuss.

-------

`rails generate react:install` is broken, I believe by [this](https://github.com/rails/webpacker/pull/636) Webpacker change. `Webpacker::Configuration` should now be `Webpacker.config`. Due to this, following the [documentation in the README](https://github.com/reactjs/react-rails/#use-with-webpacker) no longer works.

This is while using Webpacker master (new version hasn't been released yet to include these changes), which is required to [fix an issue](https://github.com/rails/webpacker/issues/673#issuecomment-323592030) with on-demand compiling in development.

### LINKED ISSUE #776: undefined method `lookup' for Webpacker::Manifest:Class

- [+] I'm trying to use `react-rails` with another library, but I'm having trouble. I've described my JavaScript management setup (eg, Sprockets, Webpack...), how I'm trying to use this other library, and why it's not working.
- [ ] I have another issue to discuss.

-------

After upgrading to Webpacker 3.0 get this error:
'undefined method `lookup' for Webpacker::Manifest:Class'

use with react-rails. Before update everything was good.

### LINKED ISSUE #778: `javascript_dir': undefined method `source_path' for Webpacker::Configuration:Class (NoMethodError)
I am trying to install 'react-rails' with 'webpacker'. I added both this gems in my rails gemfile. Versions for ruby, rails & gems are as follows:-
1. Ruby: 2.4.1
2. Rails: 5.1
3. react-rails gem: 2.2.1
4. webpacker: 3.0.0

I am not able to run "rails generate react:install" command to generate components/ directory for my React components.

I found that there was an issue in ["**react-rails-2.2.1/lib/generators/react/install_generator.rb**"](https://github.com/reactjs/react-rails/blob/master/lib/generators/react/install_generator.rb#L61) file for **javascript_dir** method.

**Earlier code:-** 
`                    def javascript_dir`
                           `if webpacker?`
                             ` Webpacker::Configuration.source_path`
                                 `.join(Webpacker::Configuration.entry_path)`
                                ` .relative_path_from(::Rails.root)`
                                 `.to_s`
                           `else`
                               ` 'app/assets/javascripts'`
                           `end`
                       `end`

**Code to fix:** 
`def javascript_dir`
`  if webpacker?`
`    Webpacker.config.source_path`
`      .join(Webpacker.config.source_entry_path)`
`      .relative_path_from(::Rails.root)`
`      .to_s`
`  else`
`    'app/assets/javascripts'`
`  end`
`end`

The term `Webpacker::Configuration` is chnaged to `Webpacker.config` and method name `entry_path` is changed to `source_entry_path`.

This work for me..!! Hope this works for you guys!

BookOfGreg:
Fixes #776 
Fixes #775 
Fixes #778
Fixes #780 
Fixes #781 
Fixes #782

Once again I could probably use some assistance with the appraisal gem but other than that should be OK.

Edit: This turned into a BIG refactor of the tests to split out support and I'll be honest my code isn't the tidiest. Still plenty of refactoring to go as I'm not sure I'm thrilled about how I've switched method definitions of Webpacker version numbers.


lib/generators/react/install_generator.rb
```diff
@@ -110,6 +110,12 @@ def setup_react_webpacker
           create_file(manifest, WEBPACKER_SETUP_UJS)
         end
       end
+
+      private
+
+      def webpack_configuration
+        Webpacker.respond_to?(:config) ? Webpacker.config : webpack_configuration
```
railsme: typo. should be
```
Webpacker.respond_to?(:config) ? Webpacker.config : Webpacker::Configuration
```

BookOfGreg: Aaw whoops. Wondered why my test didn't pass XD Thanks

BookOfGreg: I've still got issues around `Webpacker::Manifest.load`, seems to be how I detect `respond_to?` may be wrong for that one.

lib/generators/react/install_generator.rb
```diff
@@ -60,8 +60,8 @@ def webpacker?
 
       def javascript_dir
         if webpacker?
-          Webpacker::Configuration.source_path
-            .join(Webpacker::Configuration.entry_path)
+          webpack_configuration.source_path
+            .join(webpack_configuration.entry_path)
```
bradleesand: The new singleton implementation of `Webpacker::Configuration` no longer has `entry_path`. Instead, `Webpacker::Configuration.source_path.join(Webpacker::Configuration.entry_path)` becomes `Webpacker.config.source_entry_path`

vikash22292: I have suggested these changes in issue #778 .Please have a look into it. I was trying to install react with rails and failed to run react installer command. I look into gem and found that method calling is not proper. 

I fix this with minor changes in `lib/generators/react/install_generator.rb` file.

`source_path` is a instance method and was being called by class directly. Also `entry_path` was changed to `source_entry_path` in webpacker config. I have suggested these changes in #778 

The main purpose for issue was #778 was to elevate exact issue and its solution along with it, as there was no fix until that time.

basicBrogrammer: Using webpacker 3.0.1 Webpacker.config.source_entry_path worked for me 👍

BookOfGreg:
I took on the code from #778 , thank you for that one! Makes it a little easier to update.

Currently working on trying to make the new `config/webpacker.yml` and it's env files play nicely with webpacker V1's `paths.yml`. It's putting my manifest in the `webpacker.yml`'s output location with the `paths.yml`'s manifest name so the webpacker helpers can't find the manifest.

Will take me some time to figure out all the subtleties of this before I'm done. Very welcome for someone to help with this!

BookOfGreg:
Webpacker 1 and 2 have `lookup_path`, Webpacker 3 does not. and I'll have to compose a method of finding the actual path of a file.

basicBrogrammer:
I'm not sure if this is related, but the component generator is putting my components in app/javascript/packs/components. To get this to work, I had to change the require context from 'components' to './components' 
```
app/javascript/packs/application.js
var componentRequireContext = require.context('./components', true')
```
And also in my es6 components I have to require('../components/NameOfComponent') 
Any thoughts?

test/server_rendered_html_test.rb
```diff
@@ -61,15 +61,15 @@ def wait_to_ensure_asset_pipeline_detects_changes
...
+var React = require("react")
+module.exports = function() { return <span>"New List"</span> }
+HEREDOC
         else
           new_file_path = '../dummy/app/assets/javascripts/components/ZZ_NewComponent.js.jsx'
-          new_file_contents = <<~HEREDOC
-          var NewList = function() { return <span>"New List"</span> }
-          HEREDOC
+          new_file_contents = <<-HEREDOC
+var NewList = function() { return <span>"New List"</span> }
```
swrobel: Seems like you might've messed up the indentation here

bradleesand: The `<<-` syntax allows the closing `HEREDOC` to be indented but it won't strip space from the content so this works just fine.

lib/react/server_rendering/webpacker_manifest_container.rb
```diff
@@ -21,11 +21,23 @@ def find_asset(logical_path)
...
+      def manifest
+        Webpacker.respond_to?(:manifest) ? Webpacker.manifest : Webpacker::Manifest
+      end
+
+      def file_path path
+        manifest.respond_to?(:lookup_path) ? manifest.lookup_path(path) : File.join(config.output_path, manifest.lookup(path).split('/')[2..-1])
+      end
+
+      def config
+        !!defined?(Webpacker::Configuration) ? Webpacker::Configuration : Webpacker.config
```
willcosgrove: I think this needs to be

```rb
def config
  Webpacker.respond_to?(:config) ? Webpacker.config : Webpacker::Configuration
end
```

Because even in Webpacker 3, the `Webpacker::Configuration` constant is defined.

tomwaddington:
Just ran into issue #778, and this fixed it for me. 👍

BookOfGreg:
I'll admit that I'm at a bit of a dead end. On #779 I've got all of Travis passing except Rails 5 + Webpacker, but on that branch Webpack builds to public/packs but Webpacker looks in packs/packs instead. That is the case even if I peg Webpack at 1.15.x and Webpacker at ~> 1.0.
If a regular committer could help me out to get the tests passing on master then I could continue working here to get more Webpacker support.

We should really think about the maintainability of the Webpacker support as the manifest paths between versions seems really flakey and i'm not keen on the pattern of .responds_to? and defined? everywhere.

lib/generators/react/install_generator.rb
```diff
@@ -110,6 +109,16 @@ def setup_react_webpacker
           create_file(manifest, WEBPACKER_SETUP_UJS)
         end
       end
+
+      private
+
+      def webpack_source_path
+        if Webpacker.respond_to?(:config)
+          Webpacker.config.source_entry_path # Webpacker >3
```
nonmadden: it works for me
``` 
Webpacker.config.source_path.join(Webpacker.config.source_entry_path)

```

bradleesand: It works..... but is it correct?

bradleesand: I think this is correct as-is: `Webpacker.config.source_entry_path`

BookOfGreg:
Master finally passes and this does mean Webpacker 3 support (And 1.1, 1.2, 2.0) on this branch.
I split out as much as I could do make it clearer but there is still a lot of mess I caused in the code, lots of refactoring to do.

lib/react/server_rendering/webpacker_manifest_container.rb
```diff
@@ -1,4 +1,5 @@
 require "open-uri"
+# require 'pry'
```
RiccardoMargiotta: @BookOfGreg 😄

rmosolgo:
Wow, this is __really__ impressive! Thanks for all your work in digging in on this. It's nice to see that Webpacker has provided some APIs to support this kind of thing, too. 

I see the green check, and the tests look good. So, what's still to go on this one?

nonmadden:
👍

lib/generators/react/install_generator.rb
vikash22292: `Webpacker.config.source_entry_path` is the correct way to call this, as this is defined in webpacker.

BookOfGreg:
This *could* go in to master it's current state as I believe it to be an improvement but frankly this could use some TLC and refactoring.

I'm certain all my `MAJOR < 3` checks could be extracted into a class following the existing pattern.
There are probably better ways of integrating with the `Webpacker.dev_server` also.

I found some odd behaviour such as webpack 3 seems to read the wrong node_modules folder when loading in `../../../../react_ujs`, so I used the actual npm package for it instead, that might cause some people in the wild issues that I can't even predict if they're messing with node context.

All these tweaks could be in future patches to this though.

ChennyBaBy:
You beast @BookOfGreg 👍 ❗️

BookOfGreg:
I think we should probably merge this if we can be confident that it works. All existing tests pass. I can carry on with a tidy up beyond this.

BookOfGreg:
Tests fail on specific seeds:
`bundle exec appraisal rails-5_no_sprockets_webpacker_3 rake test TESTOPTS="--seed=17880"`

BookOfGreg:
@rmosolgo If you like, :shipit: and I'll keep PR'ing rubocops, test fixes and refactorings as patch-level changes after this. Be good to start getting some bug reports from folks so I know if this accidentally impacts anyone.

rmosolgo:
That sounds great, let me go ahead with a release!

gaiapunk:
I'm still having the #778  issue and was wondering if the latest merge had been release to fixed it or if there is a workaround on my end that I can do before a fix is released, thanks for all the work you do!

BookOfGreg:
It's not been released as a gem yet but it's possible to use the master branch directly with:
`gem 'react-rails', git: 'https://github.com/reactjs/react-rails.git', branch: 'master'`

PR #779, opened 2017-09-05
Get master passing again

BookOfGreg:
Master branch does not pass travis on it's own, making it difficult to tell when new branches are complete.
I plan on cleaning up the appraisal setup with the aim of making this slightly easier to work on.

- [x] Handle the Rails 5.1.0.rc1 compatibility
- [x] Handle turbolinks better
- [ ] Fix broken test on `BUNDLE_GEMFILE=/Users/e0052727/source/react-rails/gemfiles/rails_3.2.gemfile bundle exec rake test` at `test_it_reloads_when_new_jsx_files_are_added_to_the_asset_pipeline` *Seems to be a flakey test, passes sometimes?*
- [x] Consistently name tests from `test "foo bar" do` to `def test_foo_bar`
- [ ] Rubocop with the standard ruby style
- [ ]  Fix broken test on rails 5.0 + webpacker 1.2 around manifest_path


swrobel:
Why not use the `test` keyword? Rails still recommends it in their [edgeguides](http://edgeguides.rubyonrails.org/testing.html#rails-meets-minitest) and says the following: "Although you can still use regular method definitions, using the test macro allows for a more readable test name."

BookOfGreg:
When a test fails with "FAILED: ControllerTest#test_something_happened", you can global find it if the test name matches `test_something_happened` but not `test "some-thin'g ()happened"`.

Perhaps this is personal preference but it's much much faster for me to fix tests if I can find them.

BookOfGreg:
Think I found the cause of some of my remaining test failures. I'd been running Webpacker 1 against Webpack 2, which is why all my test files keep ending up in the wrong positions. Will probably implement a version check like https://github.com/shakacode/react_on_rails/blob/master/lib/react_on_rails/version_checker.rb to make at least the tests a bit more stable.

I'm a little concerned at how much work was involved at getting the master branch passing so far though. Am I doing something wrong or missing something obvious?

rmosolgo:
@BookOfGreg  Thanks digging in on this! 

> how much work

I'm _not_ surprised 😬  It was a lot of work to get webpacker working in the first place, and since then, the Webpacker APIs have changed a lot, so I bet it will be hard to get things working properly again 😖

BookOfGreg:
FYI The last fail on travis is in the `webpacker:compile` Rake task which is in `capture_io`, hence why there is no error output.
It's complaining about looking for the manifest in `packs/packs` when webpack put it in `public/packs`.
Honestly not sure what to do about that.

BookOfGreg:
Closing in favour of #777

PR #783, opened 2017-09-19
Cleanup tests, fix a broken test seed

BookOfGreg:
Hi again,

**Problem**
Noticed when 777 was merged to master, there was one timing problem test where in webpacker 3 
it's possible for it to say `Webpacker.dev_server.running? == true` before it was able to serve assets.

**Solution**
Check if the file is available in Webpacker 3 by opening it again, same as old method but using the dev_server methods. Also basically rewrote that loop while I was at it since we can just kill the PID and not need to do the `lsof` scan.

Also used the `param:` syntax for rails controller tests, it runs in Rails 3 and 5, and it also stops a deprecation warning so thats a freebie.

Also collected all the `MAJOR < 3` into one block, still not 100% sure about that patten but it seems to do a good job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
 rvm:
   - 2.3.0
   - 2.1
-  - jruby-9.0.1.0
+  - jruby-9.1.9.0
 
 gemfile:
   # These have webpacker:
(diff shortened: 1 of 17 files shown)


rmosolgo:
:+1: All sounds good to me! Thanks for taking care of the `params:` upgrade. 

Yeah, the `MAJOR < 3` thing is a bit unusual but I can't think of another way to get it done properly. 

By the way, what do you think about long-term support for webpacker 1 & 2, should we/can we drop them in the near future, or are people stuck on them?

BookOfGreg:
Can't really speak for the whole community, but if we look at the numbers I'm probably say definitely drop 1.1 (I wrote one method specific to that version), maybe 1.2 when 3 picks up.

17,079 downloads  https://rubygems.org/gems/webpacker/versions/3.0.1
3,786 downloads   https://rubygems.org/gems/webpacker/versions/3.0.0
179,128 downloads  https://rubygems.org/gems/webpacker/versions/2.0
59,125 downloads  https://rubygems.org/gems/webpacker/versions/1.2
17,494 downloads  https://rubygems.org/gems/webpacker/versions/1.1

No idea what I did to break the test suite so thoroughly but at least it looks like some JRuby builds are running now. I'll spend more time on this later, I have to do some work for my company for a few days before I can return to this.

BookOfGreg:
@rmosolgo The JRuby tests are now green also on this branch. 🚢

rmosolgo:
You're amazing!

PR #789, opened 2017-09-26
Update react to 15.6.2

BookOfGreg:
Update bundled React to 15.6.2

I know React 16 is out next week. Worth shipping this now or wait until next week anyway? Be good to have a gem release with the latest 15.x and let people update to 16.x at will.

@rmosolgo Should we cut a gem release too with all the new goodies in? I could look into 774 is that's the only oddity we've seen.


rmosolgo:
> release with the latest 15.x and let people update to 16.x at will.

totally agree!

BookOfGreg:
Thought React 16 was next week, looks like it got released just now!
I'll have to spend some time tomorrow to see why tests didn't pass just there and see if we can get a few releases out.

BookOfGreg:
`bundle exec appraisal rails-5.1-sprockets_4 rake test TESTOPTS="--seed=56135"` seems to work.
Not sure why it's failing in travis. Will bump a file to get it to re-run.

Issue #790, opened 2017-09-27
Update React to 16


BookOfGreg:
Help us help you! Please choose one:

- [ ] My app crashes with `react-rails`, so I've included the stack trace and the exact steps which make it crash.
- [ ] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.
- [ ] I'm trying to use `react-rails` with another library, but I'm having trouble. I've described my JavaScript management setup (eg, Sprockets, Webpack...), how I'm trying to use this other library, and why it's not working.
- [x] I have another issue to discuss.

-------

React 16 was released earlier.
Time to update the gem!

I'm aware the tests have some deprecation warnings so we may have to update those, and the generators may want updating too to `export default class Post extends Component` style

Example deprecation:
```
Warning: ReactTestUtils has been moved to react-dom/test-utils. Update references to remove this warning.
Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
Warning: Accessing factories like React.DOM.div has been deprecated and will be removed in v16.0+. Use the react-dom-factories package instead.  Version 1.0 provides a drop-in replacement. For more info, see https://fb.me/react-dom-factories
```

rdjpalmer:
@BookOfGreg Do you know when you plan on releasing this?

BookOfGreg:
@rdjpalmer As soon as I have some free time, I'm working on this project kindly sponsored by my day job so I have to complete other priorities, hopefully in the next couple of days.

BookOfGreg:
@rdjpalmer Shipped on my weekend, enjoy :)
Let me know if you have problems with the migration to react 16 in sprockets, I was wondering if I needed to add a teeny migration doc at all for it.

rdjpalmer:
@BookOfGreg ace, thanks for the heads up! Will let you know how it goes 👍

PR #791, opened 2017-09-27
Prepare for 2.3.0 release

BookOfGreg:
@rmosolgo Thanks for the readme additions for a release. Do we need to publish to NPM even if there is no JS change?

If so I'll get set up there and let you know the details for it, and if not then I will try a gem release.
Believe I already updated the Changelog with both PRs merged since 2.2.1 so that *should* be everything.

In this release:
- #777 Webpacker 3 support
- #789 React 15.6.2 update.

Planned next point release:
- Any bugfixes.
- #790 React 16 update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # react-rails
 
+#### Breaking Changes
+
+#### New Features
+
+#### Deprecation
+
+#### Bug Fixes
+
+## 2.3.0
+
 #### New Features
 
 - Webpacker and Webpack 3 support #777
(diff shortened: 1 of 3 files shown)


rmosolgo:
> Do we need to publish to NPM even if there is no JS change?

Great question. What I've done so far is keep the NPM version the same as the RubyGems version. But I can't think of any _real_ reason to do that. Let me see if I documented it anyplace ... 

No, I don't think so. So IMO no need to publish to NPM. 

This is gonna be sweet!

rmosolgo:
it's happening

PR #792, opened 2017-09-27
Update to react 16

### LINKED ISSUE #790: Update React to 16

BookOfGreg:
Based on [facebook's blog posts and deprecation warnings](https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html#upgrading)
- [x] Update React Builds to 16
  - [x] `react-addons-transition-group` to `react-transition-group/TransitionGroup`
  - [x] `react-addons-css-transition-group` to `react-transition-group/CSSTransitionGroup`
  - [x] `react-addons-update` to `immutability-helper`
  - [x] `react-addons-test-utils` to `react-dom/test-utils`
  - [x] Removed: `react-addons-linked-state-mixin`
  - [x] Removed: `react-addons-create-fragment`
  - [x] Removed: `react-addons-pure-render-mixin`
  - [x] Removed: `react-addons-perf`
- [x] Update generators to be compatible with 16
- [x] Find appropriate place to `require("create-react-class")` if we use that rather than ES6. In React_UJS perhaps?
  - [x] Put prop-types in bundle
  - [x] Put create-react-class in addons
- [x] Update Dummy apps in test suite
  - [x] Deprecated use of old React features
  - [x] Update to newer Libv8 (and miniracer)
  - [x] Handle "any" prop issue by removing addons

Fixes #790 

This PR is very unlikely to be backwards compatible with React 15
- Addons in the prepackaged build have changed; some have direct replacements but not all, those that don't have different API's.
- Changes like [create-class](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html) will change how our generators work.


lib/generators/react/component_generator.rb
```diff
@@ -129,15 +129,23 @@ def component_name
 
       def file_header
         if webpacker?
-          %|var React = require("react")\n|
+          if options[:es6]
```
rmosolgo: Is it important to distinguish between these two varieties of preprocessed JS? I mean, what's the scenario where you _want_ `require(...)` but not `import`? 

Anyone who's using webpacker will probably _already_ use Babel (for JSX), right? So do you think one of these options would be sufficient by itself?

BookOfGreg: I was tempted to scrap all the old style generators but was not sure on the effect it would have with sprockets + babel transpiler. I did what looked safe given I still don't fully know the system and saw the `option[:es6]` used nearby in the same file.
Will continue on this path a little this afternoon, seems like the most pressing issue for now, then maybe the HMR ( 793 ) thing after once I get time.

BookOfGreg:
@rmosolgo
The only failing tests on `bundle exec appraisal rails-5_no_sprockets_webpacker_3 rake test` are around `ExecJS::ProgramError: TypeError: Cannot read property 'any' of undefined` which I've tracked back to `addons.TransitionGroup = require("react-transition-group/TransitionGroup");` which seems to include `component.PropType.any` which is always undefined.

I couldn't find a way of getting it to accept the migration to `prop-types` so I'm considering dropping the addons packages given React 16 doesn't do addons anymore anyway.

Thoughts?

rmosolgo:
Yeah, I guess there are other ways to get addons these days, so it's not as important to bundle them here. That two-release plan sounds good, that seems like it would be easier for people to migrate, too.

BookOfGreg:
Previous version has 3k downloads. Not hearing anything about bugs in its use. Tempting to get ready to merge this one in.

rmosolgo:
Do your thing 😎

PR #796, opened 2017-09-30
Update javascript driver to headless chrome

BookOfGreg:
Been having trouble installing the correct combinations of yarn, node (7.x), and phantomJS (2.1.1) to make the tests run smoothly on other computers. Even though I have the same versions, Cliver has decided not to load Phantom on certain machines.

Decided it would make developing this gem easier for other new people if everything they needed was in the Gemspec which you can do with headless chrome. Which is also the most up to date option.
Based on: https://robots.thoughtbot.com/headless-feature-specs-with-chrome and https://docs.travis-ci.com/user/gui-and-headless-browsers/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: ruby
 sudo: required
 dist: trusty
+addons:
+  chrome: stable
 cache:
   bundler: true
   yarn: true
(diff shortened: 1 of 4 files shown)


BookOfGreg:
Wow, this actually doubles the build speed from 1h25 to 45 minutes for travis.

rmosolgo:
😍 So nice to get that download out of the CI build and put all the deps in the gemspec!

BookOfGreg:
The one 'hidden' dep is Chrome/Chromium itself. If we're happy with that (maybe I should document it) then we could merge this.

rmosolgo:
Yeah, if you add a doc, then big :+1: from me!

PR #799, opened 2017-10-03
CoffeeScript React component generator fix

ViXP:
Class with @ didn't work with export in footer (there is no need to set classes in global scope if they will be exported anyway, imho)


BookOfGreg:
Agreed.
Just spun up my first webpack + Coffeescript project [here](https://github.com/BookOfGreg/react-rails-example-app/pull/1) so that I can actually start testing these things for you. Yep you're totally right, no `@` needed.
Later on I'll write a test for this in this project but thanks.

BookOfGreg:
Added a coffee wiki entry.
https://github.com/reactjs/react-rails/wiki/React-Rails,-Coffeescript,-JSX-and-Webpacker

rmosolgo:
👏

PR #801, opened 2017-10-09
Use title case for "Webpacker" in README

djpowers:
- updates instances that refer to the project name

This change updates the casing of "Webpacker" to be consistent with how it is used elsewhere, as in this [Webpacker announcement blog post](http://weblog.rubyonrails.org/2017/8/30/webpacker-3-0/).


BookOfGreg:
Thank you!

PR #802, opened 2017-10-10
Fix Turbolinks Detection

### LINKED ISSUE #743: detectEvents() results in Turbolinks is not defined

- [x] My app doesn't crash, but I'm getting unexpected behavior. So, I've described the unexpected behavior and suggested a new behavior.
- [ ] I have another issue to discuss.

-------

I've tried this in a large react-rails app that I've been trying to upgrade to 2.0 as well as a fresh app, neither of which includes turbolinks. I was attempting to debug why `ReactRailsUJS.mountComponents()` doesn't seem to be running automatically, which is what led me to attempt to call `ReactRailsUJS.detectEvents()`. I do not see any errors in the console when my page loads, only after calling `detectEvents()`.

```javascript
> ReactRailsUJS.detectEvents()
detect.js?5263:12 Uncaught ReferenceError: Turbolinks is not defined
    at module.exports (eval at <anonymous> (application.js:3324), <anonymous>:12:5)
    at Object.detectEvents (eval at <anonymous> (application.js:3313), <anonymous>:114:5)
    at <anonymous>:1:15
```

BookOfGreg:
Fixes #743

### DIFF
--- a/lib/assets/javascripts/react_ujs.js
+++ b/lib/assets/javascripts/react_ujs.js
@@ -119,7 +119,7 @@ var turbolinksClassicEvents = __webpack_require__(10)
 module.exports = function(ujs) {
   if (ujs.handleEvent) {
     // We're calling this a second time -- remove previous handlers
-    if (typeof Turbolinks.EVENTS !== "undefined") {
+    if (typeof Turbolinks !== "undefined" && typeof Turbolinks.EVENTS !== "undefined") {
       turbolinksClassicEvents.teardown(ujs);
     }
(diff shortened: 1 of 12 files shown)


rmosolgo:
:+1: That's a nice way to test without turbolinks!

BookOfGreg:
That's why it took me way longer than I thought, tore up half the test suite figuring out why Sprockets 3 or 4 refused to load the file. There is no equivalent for Webpacker yet. I can see why it's been tricky for you, you almost want to keep both Sprockets and Webpacker separated completely for ease of development...

PR #806, opened 2017-10-12
Move package.json to top level

BookOfGreg:
This PR moves the package.json to the top level so it automatically picks up the readme and licence, to stop our NPM entry being to empty.

Not 100% it will work as intended, is there a nice way of verifying that the `npm pack` [(doc)](https://docs.npmjs.com/cli/pack) will do what we want before I break the NPM package @rmosolgo ?

This is based on the pattern of webpacker's package.

Edit: I could make a prerelease:
https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420


rmosolgo:
:+1: Makes sense to me, I see how you updated `main` and added `react_ujs` to `files`, so, seems like it'll be ok. 

If not, you can always tweak and make another quick release 😅

BookOfGreg:
Cool. I published as `npm publish --tag beta` as `2.3.0` and all seemed OK. Downloaded it into my test app and it was all happy. Will add the #802 changes in as `2.3.1`.

Just getting myself in a knot slightly with these branches, just me having teething issues is all.

I noticed in the `npm pack` step that both the source and dist folders were sent in the package, I may modify the node package in the future to be more minimal.

BookOfGreg:
https://www.npmjs.com/package/react_ujs

That looks better 👍

PR #815, opened 2017-10-24
Fix typo in REAME.md

joshgosse:


joshgosse:
Sorry, totally thought this would bootstrap a starter project or something! My bad.

BookOfGreg:
That's an interesting thought, what did you expect it to do?
Maybe there is some better expectation setting we could do in the Readme

PR #824, opened 2017-10-31
Pull request for react rails

semukas:


BookOfGreg:
@semukas Thanks for taking an interest in react-rails, going to close this as It seems accidental.

PR #830, opened 2017-11-04
Update docs

### LINKED ISSUE #825: TypeError: React.PropTypes is undefined. Migrating 2.3 -> 2.4
After upgrading from 2.3.1 to 2.4.0 my react components don't work anymore. In the console I get the following error:
`TypeError: React.PropTypes is undefined`
I've already seen that React.PropTypes has been deprecated and is now in its own module (prop-types), but downloading prop-types.min.js and including it before including react_ujs did not solve the problem. I guess the following statement needs to be added now:
`var PropTypes = require('prop-types');`

BookOfGreg:
@rmosolgo In this PR I change a lot of the Readme to try make it easier to follow the main sections and display the main benefits at the top of the README.
Also added Contributing and Code of Conduct as they are standard and can encourage best practice.

Small wording tweak in the Issues Template, it would be good to review it as almost all issues that come in are of the form "- [x] Discussion, something is wrong" and reproducing is hard.

I want to continue to add to the Wiki more How to's as I've been able to use the react-rails-example-app to close a good dozen issues where folk have just needed guidance. The issues I marked as Discussion are good candidates to close when we have a matching Wiki entry for them.

Also mentioned @renchap 's Webpacker-react in the related projects as it looks promising.

Thoughts?
Closes #825


rmosolgo:
:+1: Looks like a nice reorg! I appreciate putting the development instructions into CONTRIBUTING, that makes the readme a bit less intimidating 😆  Agreed about the issue template, I added it a while ago but I bet if it were revised, we could get more value out of it.

PR #838, opened 2017-11-13
Fix default config example in README

eins78:
The default directory for usage with webpacker is `app/javascript` (no plural)


BookOfGreg:
Yep confirmed, thanks for the correction! 🎉

Issue #832, opened 2017-11-04
ReactDOM.hydrate is not a function


victorbojica:
Hi,

Just updated the package and i'm running into this:

`index.js:93 Uncaught TypeError: ReactDOM.hydrate is not a function
    at Object.mountComponents (index.js:93)
    at HTMLDocument.ReactRailsUJS.handleMount (index.js:124)
    at HTMLDocument.dispatch (jquery.self-bd7ddd393353a8d2480a622e80342adf488fb6006d667e8b42e4c0073393abee.js?body=1:5227)
    at HTMLDocument.elemData.handle (jquery.self-bd7ddd393353a8d2480a622e80342adf488fb6006d667e8b42e4c0073393abee.js?body=1:4879)
    at Object.t.dispatch (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:6)
    at r.t.Controller.r.notifyApplicationAfterRender (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:7)
    at r.t.Controller.r.viewRendered (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:6)
    at n.t.Renderer.t.renderView (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:6)
    at n.t.SnapshotRenderer.n.render (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:6)
    at Function.t.Renderer.t.render (turbolinks.self-1d1fddf91adc38ac2045c51f0a3e05ca97d07d24d15a4dcbf705009106489e69.js?body=1:6)`

Probably related to #828 

Thanks.

BookOfGreg:
What version of React are you running? ReactDOM.hydrate is a function from v16.
`yarn upgrade react react-dom` should fix that right up for you 👍 Let me know how it goes.

xinranxiao:
@victorbojica make sure your `package.json` has this specific version for `react_usj`: `"react_ujs": "2.1.0"` (as opposed to something like `"react_ujs": "^2.1.0"`). 

[...]

PR #840, opened 2017-11-14
Prepare for 2.4.2

BookOfGreg:
### Summary

#832


BookOfGreg:
Travis is upset due to test failing on TheRubyRacer, but we removed it ages ago so no idea where it picked that gemfile up from.

Issue #842, opened 2017-11-21
Warning "Expected server HTML to contain a matching <div> in <div>"


jsbaltodano:
Well I am creating a component with 
<%= react_component('NameOfComponent', {}, prerender: false) %>

and React is logging into the console
Warning "Expected server HTML to contain a matching **div** in **div**"

For what I seen  it does not log it if the component is creating using render instead of hydrate

My question is how can I stop the warning.

I read https://github.com/reactjs/react-rails/pull/828
and then [in your code](https://github.com/reactjs/react-rails/blob/a913f6b41593e639af28cfe9d48e99012b889a38/react_ujs/dist/react_ujs.js)

```
        if (typeof ReactDOM.hydrate === "function") {
          ReactDOM.hydrate(React.createElement(constructor, props), node);
        } else {
          ReactDOM.render(React.createElement(constructor, props), node);
        }
```

and I thought, maybe you are forcing the hydrate function.

I did force the same method but without using hydrate and the warning goes away.

### System configuration
**Webpacker version 3.0.2:
**React-Rails version 2.4.1:
**Rails version 5.1.4:
**Ruby version2.4.2:

BookOfGreg:
We probably do want to be more subtle about when we hydrate vs render.
In the longer term I'd like to add a data attribute when it's pre-rendered so I can make the UJS choose between rendering and hydrating for each component. For now Hydrate is a better default than render though it does have some side effects.

I'd welcome a PR to choose `hydrate` or `render` based on if a component was pre-rendered for if anyone has time to contribute.

dwightwatson:
Spotted this one as well. We're not pre-rendering any components and this error has started popping up in the console.

phuctm4192:
I'm newbie at this gem. first of all, I installed this gem with webpacker and i also meet this issue. So how can I resolve, please tell me detail because I don't understand you said above. thanks in advanced.

jsbaltodano:
Well, it is an issue within the gem but the idea to hack it for a quick check is

Create the file component_mount.rb inside config\initializers

Write the same as the original file but you could add
`data[:hydrate] = 't' if prerender_options`

below
`data[:react_props] = (props.is_a?(String) ? props : props.to_json)`

This would work as a data param for JS to check if ReactDOM.hydrate() must be use or to use the ReactDOM.render() method.

Now in your \app\javascript\packs\application.js
`
import React from 'react';
import ReactDOM from 'react-dom';

[...]

PR #844, opened 2017-11-27
:bug: Use ReactDOM.render instead of hydrate when prerender is falsy

jsbaltodano:
Allow react-rails to use ReactDOM.render in components created whose value for prerender is falsy (false, nil).

### Summary
React gives a warning when a not pre-rendered component is not rendered using ReactDOM.render()
Following advise from #842 
A new html option has been added to use between the ReactDOM methods render() or hydrate().

render()
prerender not set or set to a falsy value

hydrate()
truthy value and not :static


BookOfGreg:
This is excellent, thanks for the contribution!
I'll make a small tweak to the changelog to reference the issue and this PR and add a ruby test for the new data attr after I merge this. Great help! 😁

PR #849, opened 2017-12-13
Modify an example in the README

ttanimichi:
We probably don't have `config/environments/application.rb` by default. Do you mean `config/application.rb` or `config/environments/production.rb` ?

```
$ rails new myapp
$ ls config/environments/
development.rb	production.rb	test.rb
```


BookOfGreg:
I did mean `config/application.rb` yes.

Thank you for your help! 💚 💙

PR #852, opened 2017-12-16
Ignore react library in code climate

Br3nda:
### Summary

code climate is enabled for this repo - so when viewing the project page, you see this:
image

and then this  claim it would take 8 months of dev time to fix the tech debt.
image

Most of these are in the vendor/react asset files.

this adds a config file telling code climate to ignore vendor files
I think I identified these correctly, but I'm unsure.

If I've configured this right, it'll mean people are no longer greeted with warnings against using this gem

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,6 @@
-exclude_paths:
-- lib/assets/javascripts/JSXTransformer.js
-- lib/assets/react-source/
+version: '2'
+exclude_patterns:
+- lib/assets/
+- react-builds/
+- react_ujs/dist/
+- test/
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Excellent! I may extend this now that you have me reading up on it.
My reading list:
https://docs.codeclimate.com/v1.0/docs/advanced-configuration
https://docs.codeclimate.com/v1.0/docs/excluding-files-and-folders

Thanks @Br3nda

BookOfGreg:
@Br3nda Tweaked your PR slightly to use V2 of CodeClimate config based on reading the docs your PR led me to.
Thanks a bunch! Learned a load diving into there. Will merge when it runs green and see how our CodeClimate looks.

BookOfGreg:
Thank you @Br3nda !

screen shot 2017-12-19 at 10 20 24

Issue #850, opened 2017-12-14
Generators should use Fragments instead of Divs


BookOfGreg:
### Expected behavior
Generator should make components with Fragments by default from React v16.2.0

https://reactjs.org/docs/fragments.html
https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

### Actual behavior
Uses Div by default

PR #856, opened 2017-12-27
🔥  use Fragments instead of Divs

rahulcs:
### Summary

Allow react-rails to use React.Fragment in components created instead of Divs. implements #850 

### Other Information

also updates React version to 16.2.0


BookOfGreg:
😍 Lovely! Thank you for the contribution, it really helps 👍 
Have a happy new year.

rahulcs:
@BookOfGreg Thanks for maintaining this! You have a happy new year as well!

PR #857, opened 2017-12-31
CI against Ruby 2.5.0

ttanimichi:


BookOfGreg:
Thanks for your your PR.
Some versions of Rails do not appear to be compatible with Ruby 2.5 so they will need to be commented out.

ttanimichi:
@BookOfGreg done https://github.com/reactjs/react-rails/pull/857/commits/9b1019b8ceea0264b853d7768b06cc2082e6060f

PR #862, opened 2018-01-16
Fix misleading comment in exec_js_renderer

not-an-aardvark:
### Summary

Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

As of 41721c5374925384adc8679aa007789b598de969, `exec_js_renderer` now needs `ReactRailsUJS` to appear in the global context, not `ReactDOM`. This commit updates an outdated comment to avoid confusion.

### DIFF
--- a/lib/react/server_rendering/exec_js_renderer.rb
+++ b/lib/react/server_rendering/exec_js_renderer.rb
@@ -1,7 +1,7 @@
 module React
   module ServerRendering
     # A bare-bones renderer for React.js + Exec.js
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Thank you for the update and the research into when it broke, nice detective work!

Looks like Travis broke while building your PR but that doesn't matter as this one is just docs. Still kicked it off again by hand to get the green dot for historical use.

Will merge once green 👍

PR #865, opened 2018-01-26
fix(README): correct react:install for rails < 5.0

codef0rmer:
### Corrected installation commands for Rails version < 5.0

These commands will not work on rails < 5.0

```
$ rails webpacker:install
$ rails webpacker:install:react
```


BookOfGreg:
Thanks @codef0rmer , and using semantic commit message format too 🎉

PR #866, opened 2018-02-04
chore: :fire: Remove unsupported ruby and rails versions

BookOfGreg:
### Summary

Our two main integrations do not support things indefinitely. To encourage people to move to secure and updated versions, we will no longer run our tests on those platforms. If our gem continues to run on Rails older than 4.2, it will be incidental from here on.

http://guides.rubyonrails.org/maintenance_policy.html
https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

### Other Information

This will drop explicit support for Ruby 2.1 and Rails 3.2, 4.0, 4.1.

This may also make the Travis run a little faster.


rmosolgo:
👏 !

PR #869, opened 2018-02-07
Controller camelize_props: true

ttanimichi:
This allows the use of camelize props in controller actions.

```ruby
render component: 'Example', props: my_props, camelize_props: true
```


BookOfGreg:
Hi @ttanimichi 
:+1: thanks for the fix and for going through the time to update the tests!
Interesting that the controller_renderer and the component_mount handle their args differently. I'll keep that in mind for an eventual refactor there.

For my notes:
https://github.com/reactjs/react-rails/pull/409
https://github.com/reactjs/react-rails/blob/master/lib/react/rails/component_mount.rb
https://github.com/reactjs/react-rails/blob/master/lib/react/rails/controller_renderer.rb

PR #873, opened 2018-02-15
:memo: Add Typescript integration details

iporollo:
### Summary

Added a Typescript integration guide to README.md based on https://github.com/rails/webpacker/blob/master/docs/typescript.md


BookOfGreg:
<3 Lovely. And thanks for using the tags and emojis!
I'll probably link the original text in a later commit but thanks 😄

Issue #892, opened 2018-04-16
One of the test suite deps has a regression


BookOfGreg:
### Steps to reproduce

Look at the Travis

### Expected behaviour
Green 🎉 

### Actual behaviour
Red 😢 

Last successful one was https://travis-ci.org/reactjs/react-rails/jobs/342097144
Check the bundle section and play a game of whack-a-gem. Can replicate locally if I `bundle exec appraisal update` then `bundle exec appraisal rake`.

This was not caught as my appraisal install was stale.

suhailnaw:
I'd like to take a stab at this!  I'm looking to make my first open source contribution and will take a few hours to look at it this week.

BookOfGreg:
@suhailnaw Thank you, it's certainly possible to simply peg the version of all the gems down for that appraisal to solve it, but I'd like to know which specific gem had a backward-incompatible change. (So we can go blame them 😉  )

Good luck and most importantly have fun.

suhailnaw:
Hey @BookOfGreg I'm still working on this issue.  I wanted to update you on my work so far.

I compared the versions of the Ruby gems from the broken builds and the working builds and did not find a difference.  I then thought it might be the version of Chrome that was the issue, but changing it in the PR above did not work.

My next step is to set up the local dev environment on my computer.  I'm going to use rbenv to manage the two versions of Ruby we're interested in, 2.3.0 and 2.5.0, and try to get the builds to work.  Do you have any ideas of where to look?  

I would also like to use TravisCI to troubleshoot since I'm finding the graphical interface of build history useful.  But I don't want to crowd your commit history.  Is the standard procedure just to develop locally or can I use TravisCI?

BookOfGreg:
I suspected the gems because it's not easy to get a [gemfile lock](https://github.com/reactjs/react-rails/tree/master/gemfiles) when using [appraisal](https://github.com/thoughtbot/appraisal)

Additionally I was able to run the tests on my local machine but when I removed my appraisal lockfiles it failed in the same way.
Perhaps there is a dependency in the appraisals that should be locked lower but isn't. I already tried webpacker as they tend to break stuff and move quickly but wasn't them as far as I can tell this time.

ttanimichi:
👀

ttanimichi:
The cause of the error seems to be an incompatibility of webpacker. ref. #916

BookOfGreg:
[...]

PR #894, opened 2018-05-14
Test to fix issue #892 with chrome version change from stable to beta

suhailnaw:
### Summary

Trying to determine what causes 3 of 21 builds to fail, per Issue #892.  I already compared the previous working build's gem versions to the broken build and did not find any differences.  

The new version of Chrome (from version 64.0.3282.167-1 to 66.0.3359.139-1) might have affected the build, so I want to try to change the Chrome version.  It's installed in the travis.yml file, but you're only allowed to specify stable or beta as the version, not a number version of chrome.  

I'm creating this pull request because I want to test this change within a Travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 sudo: required
 dist: trusty
 addons:
-  chrome: stable
+  chrome: beta
 cache:
   bundler: true
   yarn: true


BookOfGreg:
In order to try help diagnosis, I just asked travis to re-run previously passing build 1417.4
https://travis-ci.org/reactjs/react-rails/builds/338407498
Not sure if a re-run will upgrade the chrome, but if it does and fails then it's chrome.

PR #900, opened 2018-06-06
📝 Document integrating with Turbolinks when imported with Webpacker

### LINKED ISSUE #899: Webpacker + Turbolinks + ReactRailsUJS = 😞
### Steps to reproduce

Create a new Rails app that uses Webpacker for assets, and **`import`s Turbolinks** (as opposed to using Sprockets for Turbolinks as seems to be the case for this repo's tests).

Run `ReactRailsUJS.detectEvents()` to set up Turbolinks' events.

### Expected behavior
`ReactRailsUJS` should `mount` and `unmount` React components on the corresponding Turbolinks events.

### Actual behavior
Those events don't get fired.

### System configuration
**Webpacker**: 3.2.0
**React-Rails version**: 2.4.4
**React_UJS version**: 2.4.4

The reason this happens is because React-Rails expects Turbolinks to be a **global** object. If you're using Webpacker, by default everything is **local**, causing React-Rails to think that Turbolinks is not installed.

(My attempt to solve this via using a `webpack.ProvidePlugin` to turn Turbolinks into a global did not work, because `Turbolinks.start()` needs to be called before React-Rails, else `Turbolinks.controller` is `undefined`.)

I'm not sure this can be considered a bug, so instead I'll just share how I solved this. If this is intended, perhaps it can be added to README / Wiki? (I can create a PR if desired!)

My solution:

```js
import Turbolinks from "turbolinks";
import ReactRailsUJS from "react_ujs";

// This needs to be in this particular order:
Turbolinks.start();
// Add Turbolinks to the global namespace
window.Turbolinks = Turbolinks;
// Remove previous (native) events, and add Turbolinks'
ReactRailsUJS.detectEvents();
// (Optional) Clean up global namespace
delete window.Turbolinks;
```

Cheers!

mendelk:
[ci skip]

resolves #899

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,19 @@ ReactRailsUJS.detectEvents()
 
 For example, if `Turbolinks` is loaded _after_ `ReactRailsUJS`, you'll need to call this again. This function removes previous handlers before adding new ones, so it's safe to call as often as needed.
 
+If `Turbolinks` is `import`ed via Webpacker (and thus not available globally), `ReactRailsUJS` will be unable to locate it. To fix this, you can temporarily add it to the global namespace:
+
+```js
+// Order is particular. First start Turbolinks:
+Turbolinks.start();
+// Add Turbolinks to the global namespace:
+window.Turbolinks = Turbolinks;
+// Remove previous event handlers and add new ones:
+ReactRailsUJS.detectEvents();
+// (Optional) Clean up global namespace:
+delete window.Turbolinks;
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Thank for for adding that to our docs, this will hopefully help others using Turbolinks in Webpacker. We will monitor to see if any other issues are opened in relation to this to guide us on if an automated solution is needed.

PR #911, opened 2018-07-23
Remove male pronoun in code comment

alextaylor000:
### Summary

Just a small tweak to a code comment to make the codebase more inclusive 😊


BookOfGreg:
Looks like that was the only instance of it according to [GH search](https://github.com/reactjs/react-rails/search?q=his&unscoped_q=his)

Thanks for contributing :)

PR #932, opened 2018-09-06
Support camelizing ActionController::Parameters

### LINKED ISSUE #931: Add support for camel-casing ActionController::Parameters
### Steps to reproduce

```ruby
class MyController < ApplicationController

  def foo
    props = params.permit(:foo_bar, :baz_etc)
    render component: 'foo', props: props, camelize_props: true
  end
end
```

Note that the given props are not properly camelized when rendered to the React component. This is because [`React.camelize_props`](https://github.com/reactjs/react-rails/blob/master/lib/react.rb) does not recognize `ActionController::Parameters` as a valid hash and returns it as-is.

### Expected behavior
`React.camelize_props` should camelize instances of `ActionController::Parameters`.

### Actual behavior
`React.camelize_props` does not camelize instances of `ActionController::Parameters`.

-------

I was going to submit a PR, but I noticed that the gem spec only requires Rails 3.2, so I wasn't sure what the best approach might be for testing with a class that didn't appear until Rails 4.

jcarlson:
### Summary

Adds support for React.camelize_props to camelize an instance of
ActionController::Parameters hash, so that permitted, strong parameters
in a Rails controller can be passed to a React component.

### Other Information

Resolves #931

### DIFF
--- a/test/react_test.rb
+++ b/test/react_test.rb
@@ -28,4 +28,23 @@ class ReactTest < ActiveSupport::TestCase
 
     assert_equal expected_props, React.camelize_props(raw_props)
   end
+
+  def test_it_camelizes_params
+    raw_params = ActionController::Parameters.new({
+      foo_bar_baz: 'foo bar baz',
+      nested_keys: {
+        qux_etc: 'bish bash bosh'
+      }
+    })
+    permitted_params = raw_params.permit(:foo_bar_baz, nested_keys: :qux_etc)
+
+    expected_params = {
+      'fooBarBaz' => 'foo bar baz',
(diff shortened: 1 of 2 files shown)


test/react_test.rb
```diff
@@ -28,4 +28,23 @@ def test_it_camelizes_props
 
     assert_equal expected_props, React.camelize_props(raw_props)
   end
+
+  def test_it_camelizes_params
```
BookOfGreg: You can actually wrap the test in a defined check like this:

```
if defined?(ActionController::Parameters)
  def test_method
  end
end
```

Which should make it work on Rails <4

PR #937, opened 2018-10-03
Use .keep instead of .gitkeep to match Rails itself.

pupeno:
Also, if someone is not using git, a .gitkeep feels odd, where a .keep seems fine.


BookOfGreg:
Thanks for the contribution, yep agree on matching Rails on this one. :+1:

PR #941, opened 2018-10-06
Add semi-colons to generated JavaScript

benbristow:
### Summary

Add semi-colons to generated JavaScript which is created when using the generator.

Seems like good practice and I always end up doing this myself after running the generator anyway for the sake of cleanliness. May as well fix it at the source, eh?

### Other Information

N/A

### DIFF
--- a/react-rails.gemspec
+++ b/react-rails.gemspec
@@ -16,19 +16,18 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'appraisal'
   s.add_development_dependency 'bundler', '>= 1.2.2'
-  s.add_development_dependency 'capybara'
-  s.add_development_dependency 'chromedriver-helper'
   s.add_development_dependency 'codeclimate-test-reporter'
   s.add_development_dependency 'coffee-rails'
   s.add_development_dependency 'es5-shim-rails', '>= 2.0.5'
-  s.add_development_dependency 'guard-minitest'
   s.add_development_dependency 'guard'
+  s.add_development_dependency 'guard-minitest'
   s.add_development_dependency 'jbuilder'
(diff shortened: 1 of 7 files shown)


react-rails.gemspec
```diff
@@ -16,18 +16,19 @@ Gem::Specification.new do |s|
...
+  s.add_development_dependency 'guard'
   s.add_development_dependency 'jbuilder'
   s.add_development_dependency 'listen', '~> 3.0.0' # support Ruby 2.1
-  s.add_development_dependency 'chromedriver-helper'
-  s.add_development_dependency 'capybara'
+  s.add_development_dependency 'rails', '>= 3.2'
   s.add_development_dependency 'selenium-webdriver'
   s.add_development_dependency 'test-unit', '~> 2.5'
-  s.add_development_dependency 'rails', '>= 3.2'
+  s.add_development_dependency 'turbolinks'
```
BookOfGreg: This file probably wants to be reset, turbolinks isn't a dependency of this project.

BookOfGreg: I see your commit eeef8a3 mentions tests failing without turbolinks but the build before yours ran green. What are you referring to?

benbristow: @BookOfGreg The first time I ran the tests on my machine (I assume they're ran with a simple `rake` ?) I was getting tons of errors about not being able to resolve Turbolinks on the controller tests.

I can roll it back if you want.

BookOfGreg: Yes please. The tests that matter for this project are the ones on Travis. Rake should work locally when using Appraise but it's easy to pollute the environment on this project accidentally.

benbristow:
@BookOfGreg reverted.

BookOfGreg:
Thank you Ben 👍

PR #946, opened 2018-10-18
Camelize all props that can be converted to JSON

j-clark:
### Summary

This solves the case where we want to camelize_props, but those props aren't explicitly a hash or array (for example `ActiveModel::Serializer`, `Representable`, `ActionController::Parameters`, etc.

As mentioned about, I'm making the assumption that ActiveSupport is loaded since this is within Rails, but let me know if that's not a valid assumption

Also, thoughts on simplifying it even further? If we know that ActiveSupport is defined, we could just use `#deep_transform_keys` as so

```ruby
def self.camelize_props(props)
  props_as_json = props.as_json

  case props_as_json
  when Hash
    props_as_json.deep_transform_keys { |key| key.to_s.camelize(:lower) }
  when Array
    props.map { |prop| camelize_props(props) }
  else
    props
  end
end

### DIFF
--- a/lib/react.rb
+++ b/lib/react.rb
@@ -3,21 +3,19 @@ module React
   # @param props [Object] If it's a Hash or Array, it will be recursed. Otherwise it will be returned.
   # @return [Hash] a new hash whose keys are camelized strings
   def self.camelize_props(props)
-    case props
+    props_as_json = props.as_json
+
+    case props_as_json
     when Hash
-      props.each_with_object({}) do |(key, value), new_props|
+      props_as_json.each_with_object({}) do |(key, value), new_props|
         new_key = key.to_s.camelize(:lower)
         new_value = camelize_props(value)
         new_props[new_key] = new_value
       end
     when Array
-      props.map { |item| camelize_props(item) }
+      props_as_json.map { |item| camelize_props(item) }
     else
-      if defined?(ActionController::Parameters) && props.is_a?(ActionController::Parameters)
-        camelize_props(props.to_h)
-      else
-        props
-      end
+      props_as_json
     end
   end
 end
--- a/test/react_test.rb
+++ b/test/react_test.rb
@@ -23,7 +23,7 @@ class ReactTest < ActiveSupport::TestCase
           [{ 'nestedArray' => {} }]
         ]
       },
-      'alreadyCamelized' => :ok
+      'alreadyCamelized' => 'ok'
     }
 
     assert_equal expected_props, React.camelize_props(raw_props)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by j-clark** [comment]
Any opinions on this?

**Comment by BookOfGreg** [comment]
It's a good upgrade over the existing params check.
Was thinking if there were any negative side effects of calling `as_json` on everything but can't think of one, all major Ruby and Rails enumerable objects likely to be passed in do declare that method as far as I know.

The doc for that methods needs to be updated to reflect this change and I need to think if this classes as a breaking change or if I can put it in a feature level patch.

Good contribution, glad that GitHub added the bookmark notifications tab as to be honest this one got forgotten about a little (sorry!)

**Comment by j-clark** [comment]
ah, good call on version. as far as i can tell, it looks like `deep_transform_keys` was added in rails 4. i think `as_json` goes back to 2.3 (according to [apidock](https://apidock.com/rails/ActiveResource/Base/as_json)). i don't know if they were necessarily added at the same time, but i'm pretty sure that anything responding to `to_json` (which is already called in the view helper) will also respond to `as_json`

i think putting this in a patch version is probably too risky. someone is inevitable passing in active record objects, which would get serialized as json correctly, but not camelized. then getting this update would likely break their apps. and this is kind of tricky to track down.

i'll update the docs later today. is it just the README? 

thanks!

File: lib/react.rb [hunk 1]
```diff
@@ -3,21 +3,19 @@ module React
   # @param props [Object] If it's a Hash or Array, it will be recursed. Otherwise it will be returned.
```
BookOfGreg: Inline docs here.

BookOfGreg:
Marked the inline doc as a comment.
Additions to the readme and/or wiki are always very welcome as they do talk about camelization in general which would be changing.

I was wondering if there was some sort of option we could put this behind to make it opt-in, thereby letting us release as a feature without breaking people's existing React apps. However since there's already code there to specifically change Params, any toggle ignoring params would be a very odd toggle indeed.

Likely at this point to make it hold off for a major version bump. Need to update pre-bundled react anyway 👍

justin808:
I think this is OK to merge as-is.

@BookOfGreg @j-clark any thoughts on this?

@j-clark can you please add a changelog entry?

justin808 (approved):
LGTM

alkesh26:
We need to CHANGELOG entry, and the PR is good to go.

j-clark:
wow, this PR is a relic from the past! i'll be honest, it's unlikely that i'll get a changelog entry added any time soon. i'm a corporate drone these days and don't even have git setup on my work laptop 😞

Judahmeek:
@justin808 I created a separate PR with the changelog entry for this PR: #1207

Let's get this merged.

Issue #947, opened 2018-10-18
Server side rendering fails with identifier 'Set' undefined


HorizonShadow:
### Steps to reproduce

* Make any old component
* Set prerender to true
* Try to render the component

### Expected behavior

It should pre-render the component.

### Actual behavior

Execjs crashes with the following error:

```  
ExecJS::ProgramError (identifier 'Set' undefined):
  
(execjs):1
app/controllers/entry_point_controller.rb:3:in `index'
```

### System configuration
Sprockets or Webpacker version: webpacker (3.5.5)
React-Rails version: react-rails (2.4.7)
React_UJS version: "react_ujs": "^2.4.4"
Rails version: rails (5.2.1)
Ruby version: RUBY VERSION: 2.4.4 (2018-03-28 patchlevel 296) [i386-mingw32]
-------

Basically everything works perfectly fine, but if I attempt to use server-side rendering, the program crashes with the above error: `identifier 'Set' undefined`. I thought it might have been due to an outdated, but I updated to the latest node and the issue persists.

BookOfGreg:
Marked invalid as your provided reproduction does not cause the effect you say for me.

Could you please provide code that exhibits the issue you are seeing?

HorizonShadow:
Yes, with prerender set to false, everything runs correctly. 

I'll create a new project and detail everything I do.

* `rails new react-rails-prerender-test --webpack=react --skip --database=postgresql`
* `bundle add react-rails`
* `rails g react:install`
* `rails g controller entry index -s`
* `rails g react:component Entry -s`
* Changed the render method of Entry.js to say "Hello World!"
* Added `    render_component :Entry` to the index route of EntryController
* `rails server -b 127.0.0.1 -p 3000 -e development`
* Navigate to localhost:3000/entry/index
```
Started GET "/entry/index" for 127.0.0.1 at 2018-10-19 07:02:03 -0300
Processing by EntryController#index as HTML
[Webpacker] Compilingâ€¦
[Webpacker] Compiled all packs in C:/Users/horiz/RubymineProjects/react-rails-prerender-test/public/packs
Completed 500 Internal Server Error in 3005ms (ActiveRecord: 0.0ms)


  
ExecJS::ProgramError (identifier 'Set' undefined):
  
(execjs):1
app/controllers/entry_controller.rb:3:in `index'
```
![image](https://user-images.githubusercontent.com/1729810/47211900-21fe2580-d36d-11e8-8ac9-de0cba32d986.png)

[...]

PR #949, opened 2018-10-24
Warn on old js runtimes

### LINKED ISSUE #948: On certain JS runtimes, send warning notice.
Related to #947 and any other issue mentioning `Set`, Node < 6 or RubyRacer.

All of those issues are due to old JS runtimes.
I would welcome a PR that detects invalid JS runtimes and warns the user to update system Node or use a libv8 gem that is up to date.

Relates to https://github.com/rubygems/rubygems/issues/1811 as having miniracer be an optional dependency would aid this effort.

I would also welcome a PR that updates the readme to document the JS runtimes.

HeroProtagonist:
Closes #948

### Summary

Added a check to see what node version is being used in `railtie.rb`. I am not a ruby/rails dev, but after some looking this seemed like an early spot to run this check. I made a custom color function as opposed to adding a library like [this](https://github.com/fazibear/colorize) one I found.

I also matched the engines in the `package.json` to those in [webpacker](https://github.com/rails/webpacker/blob/master/package.json#L10-L12)

### Other Information

A concern I would have is if this library is expected to work on non *nix environments. If that is the case then the shell commands I wrote would not work. In the node world, I would use a library like [shelljs](https://github.com/shelljs/shelljs) to handle multiple operating systems. Not sure about how this is handled in ruby.

Please let me know if that is a problem and I will fix it.

Thanks!

### DIFF
--- a/lib/react/rails/railtie.rb
+++ b/lib/react/rails/railtie.rb
@@ -1,8 +1,11 @@
 require 'rails'
+require_relative 'js_runtime_checker'
 
 module React
   module Rails
     class Railtie < ::Rails::Railtie
+      JsRuntimeChecker.call
+
       config.react = ActiveSupport::OrderedOptions.new
       # Sensible defaults. Can be overridden in application.rb
       config.react.variant = (::Rails.env.production? ? :production : :development)
@@ -97,20 +100,6 @@ module React
       config.after_initialize do |app|
         # The class isn't accessible in the configure block, so assign it here if it wasn't overridden:
         app.config.react.server_renderer ||= React::ServerRendering::BundleRenderer
-
-        def yellow(message)
-           "\e[33m#{message}\e[0m"
-        end
-
-        node_version = `echo $(node -v | cut -c 2-)`.strip
-        node_major_version = node_version.sub(/\..*/, '').to_i
-
-        if node_major_version < 6
-          puts yellow("Warning: You are using node #{node_version}.")
-          puts yellow("This an unsupported JavaScript runtime.")
-          puts yellow("Please upgrade your node version to one >=6.")
-          puts yellow("For more information see this issue https://git.io/fxXgI")
-        end
 
         React::ServerRendering.pool_size        = app.config.react.server_renderer_pool_size
         React::ServerRendering.pool_timeout     = app.config.react.server_renderer_timeout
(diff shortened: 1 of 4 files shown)


lib/react/rails/railtie.rb
```diff
@@ -97,6 +97,20 @@ class Railtie < ::Rails::Railtie
         # The class isn't accessible in the configure block, so assign it here if it wasn't overridden:
         app.config.react.server_renderer ||= React::ServerRendering::BundleRenderer
 
+        def yellow(message)
+           "\e[33m#{message}\e[0m"
+        end
+
+        node_version = `echo $(node -v | cut -c 2-)`.strip
```
BookOfGreg: We should avoid calling out to shell where possible.
ExecJS is already a dependency of ours and can abstract this cross-os for us.

See comment on checking JS runtime;
https://github.com/reactjs/react-rails/issues/947#issuecomment-431569700
Also see code on ExecJS Runtime abstraction:
https://github.com/rails/execjs/blob/master/lib/execjs/runtimes.rb

In all honesty I'm tempted to see if they'd accept a PR that would give each runtime either versions or capabilities, which we could make use of.

HeroProtagonist: Thanks for the review! I will address the comments later today and update

BookOfGreg: How's it going?

BookOfGreg:
Todo
- [ ] Update to not call shell

HeroProtagonist:
@BookOfGreg Wow this one got lost in the sauce. Coincidently, I now work in ruby/rails at my current job. I will see this one to the end. Hope you are staying safe!

BookOfGreg:
These things do get lost but it's not a big deal. I use these projects in waves too, so I also go a few months without checking in on them depending on the project I'm on.
Hope you are well!

HeroProtagonist:
Updated the PR; however, travis is failing.
Looks like it is because [master](https://travis-ci.org/github/reactjs/react-rails/builds/686187189) build is broken

BookOfGreg:
Yep, annoyingly this project is very exposed to the JavaScript treadmill so anytime we're prepping for a release we have to sort all the JS deps all over again.
Thanks for your work, I'll take a look at this one for the next release, not sure when that will be though.

alkesh26:
@justin808, We can close this PR. The file `lib/react/rails/js_runtime_checker.rb` was removed.

PR #956, opened 2018-12-27
Add gem installation to Sprocket instructions.

joshuapinter:
Skips this step that exists on the Webpacker instructions.


BookOfGreg:
Thanks for the contribution. Always useful to help the docs and make it easier to use :)

joshuapinter:
👍 Thanks for the great gem. Got React running in no time on a 8 year old legacy Rails 4.2 app. :)

PR #957, opened 2019-01-02
Add assert_react_component method for test react_component render result

huacnlee:
```rb
test 'assert_react_component' do
  get "/pages/1"
  assert_equal 200, response.status
  assert_react_component "GreetingMessage"
  assert_react_component "GreetingMessage" do |props|
    assert_equal "Bob", props[:name]
    assert_equal "Last Bob", props[:last_name]
    assert_equal "Bob", props[:info][:name]

    assert_select "[id=?]", "component"
    assert_select "[class=?]", "greeting-message"
  end
end
```

### DIFF
--- a/lib/react-rails.rb
+++ b/lib/react-rails.rb
@@ -2,3 +2,9 @@ require 'react'
 require 'react/jsx'
 require 'react/rails'
 require 'react/server_rendering'
+
+module React
+  module Rails
+    autoload :TestHelper, 'react/rails/test_helper'
+  end
+end
--- a/lib/react/rails/test_helper.rb
+++ b/lib/react/rails/test_helper.rb
@@ -1,0 +1,26 @@
+module React
+  module Rails
+    module TestHelper
+      extend ActiveSupport::Concern
+
+      # assert react_component render
+      #
+      # assert_react_component("HelloWorld") do |props|
+      #   assert_equal "Hello world", props[:message]
+      # end
+      def assert_react_component(name)
+        assert_select "div[data-react-class]" do |dom|
+          assert_select "[data-react-class=?]", name
+
+          if block_given?
+            props = JSON.parse(dom.attr("data-react-props"))
+            props.deep_transform_keys! { |key| key.to_s.underscore }
+            props.deep_symbolize_keys!
+
+            yield(props)
+          end
+        end
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -159,6 +159,35 @@ $ yarn add @types/react @types/react-dom
 
 Doing this will allow React-Rails to support the .tsx extension.
 
+### Test component
+
+You can use `assert_react_component` to test component render:
+
+app/views/welcome/index.html.erb
+
+```erb
+<%= react_component("HelloWorld", { greeting: "Hello from react-rails.", info: { name: "react-rails" } }, { class: "hello-world" }) %>
+```
+
+```rb
+class WelcomeControllerTest < ActionDispatch::IntegrationTest
+  test 'assert_react_component' do
+    get "/welcome"
+    assert_equal 200, response.status
+
+    # assert rendered react component and check the props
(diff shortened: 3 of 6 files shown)


lib/react/rails/test_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+      #   assert_equal "Hello world", props[:message]
+      # end
+      def assert_react_component(name)
+        assert_select "div[data-react-class]" do |dom|
+          assert_select "[data-react-class=?]", name
+          props = JSON.parse(dom.attr("data-react-props"))
+          props.deep_transform_keys! { |key| key.to_s.underscore }
+          props.deep_symbolize_keys!
+
+          yield(props) if block_given?
```
BookOfGreg: Since props are only used in the block, the trailing `if` could save some computation by doing the following:
```
if block_given?
  props = JSON.parse(dom.attr("data-react-props"))
  props.deep_transform_keys! { |key| key.to_s.underscore }
  props.deep_symbolize_keys!

  yield(props)
end
```

lib/react/rails/test_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Rails
+    module TestHelper
+      extend ActiveSupport::Concern
+
+      # assert react_component render
+      #
+      # assert_react_component("HelloWorld") do |props|
+      #   assert_equal "Hello world", props[:message]
+      # end
+      def assert_react_component(name)
```
BookOfGreg: Would it be useful and/or possible to also allow assertion if it is prerendered or not?

lib/react-rails.rb
```diff
@@ -2,3 +2,5 @@
 require 'react/jsx'
 require 'react/rails'
 require 'react/server_rendering'
+
+ActionDispatch::IntegrationTest.send(:include, React::Rails::TestHelper)
```
BookOfGreg: I think there's a better way of including modules in rails without using the `send` keyword, though I've not used it much myself; Devise does it in this way:
https://github.com/plataformatec/devise/blob/master/lib/devise/rails.rb#L21
and
https://github.com/plataformatec/devise/blob/master/lib/devise.rb#L12

lib/react/rails/test_helper.rb
huacnlee: This my first use react-rails and I am not use `prerender` feature, so I don't know the render logic, would you please add `prerender` support after the PR merged?

lib/react-rails.rb
huacnlee: I just updated for require test_helper by autoload, include helper by use `ActiveSupport.on_load`

huacnlee:
@BookOfGreg `ActiveSupport.on_load(:action_dispatch_integration_test)` was coming from Rails 5.0.0, so 4.2 not support that.

See here example of ActionMailer:

https://github.com/rails/rails/commit/9a642931fbcc4bab0edffe41d3e7bf59e250e90e

How a about here change to:

```rb
ActiveSupport.on_load(:action_view) do
  include ::React::Rails::ViewHelper
  ActionDispatch::IntegrationTest.send :include, React::Rails::TestHelper
end
```

BookOfGreg:
If it's good enough for Rails, it's good enough for me :) I'll accept that suggestion.

huacnlee:
@BookOfGreg just updated

BookOfGreg:
I'll clean up whatever is wrong with the test suite. Thank you for the contribution. I'll try get a release out in the coming weeks.

cacods:
Is it possible to use `assert_react_components` in sub components of the main component called with `render_component` helper?

eelkevdbos:
@cacods, I expect you need a feature test for this because it would involve actually initializing the server rendered dom elements into full react components.

The current assert_react_components helper does only asserts the existence of a `div[data-react-class=?]` with the expected properties through inspection of the `data-react-props` attribute.

PR #962, opened 2019-02-01
Enhanced turbolinks support

### LINKED ISSUE #960: Components aren't mounted on turbolinks cache
Help us help you! Have you looked for similar issues? Do you have reproduction steps? [Contributing Guide](CONTRIBUTING.md#reporting-bugs)

### Steps to reproduce

Place a component on your page. Navigation to that page, then to another, then back to the page with the component.

In the period between turbolinks showing you the cached page, and fetching the new one, no components are mounted.

### Expected behavior
The cached page should at least *show* the component. If remounting it temporarily has too many side effect. I know stimulus just mounts and unmounts it as the cache is replaced.

### Actual behavior
The components don't show up at all. There's just empty space.

### System configuration
**Sprockets or Webpacker version**: 3.5
**React-Rails version**: 2.4
**Rect_UJS version**: 2.4.4
**Rails version**: 5.2.2
**Ruby version**: 2.4.5

-------

I'm not actually sure if I should be posting this here or over at turbolinks, but seeing as not even `data-turbolinks-permanent`, I feel like it must be something react-rails is doing.

If you render a component, navigate away, and navigate back, the component is gone until turbolinks finishes loading. Even if the component has a data-turbolinks-permanent tag, or if its parent has it. The component just doesn't exist.

So if I have a slow page load, I end up with many seconds of elements just... not there. It's not very pretty.

Here's a gif of what I mean. After navigation back to the first page, the table doesn't mount.

1oumczyorq

HorizonShadow:
### Summary

This does 2 things. 

The first is move the mounting to `turbolinks:before-render`, since that's the earliest event where the user sees content on the page. When it was set to `load` only, the user wouldn't see any react components until the network request finished, rather than seeing them on the cached page. This fixes #960 

The second caches rendered components. If you pass an existing component to ReactDOM, it will update the currently mounted component with the passed component, rather than replacing it.
This allows the `data-turbolinks-permanent` attribute to work correct. When the cached page is displayed and react_ujs mounts the existing component, it won't destroy the cached component.

This lets you do things like have react components in your layout that aren't affected by page changes, like a navigation component, or search popup. 

An example where the navbar/appbar is permanent, but the main page is not:

mk2aly1ba2

### DIFF
--- a/lib/react/rails/component_mount.rb
+++ b/lib/react/rails/component_mount.rb
@@ -15,6 +15,7 @@ module React
       # You can use them in custom helper implementations
       def setup(controller)
         @controller = controller
+        @cache_ids = []
       end
 
       def teardown(controller)
@@ -40,6 +41,9 @@ module React
             data[:react_class] = name
             data[:react_props] = (props.is_a?(String) ? props : props.to_json)
             data[:hydrate] = 't' if prerender_options
+
+            num_components = @cache_ids.count { |c| c.start_with? name }
+            data[:react_cache_id] = "#{name}-#{num_components}"
           end
         end
         html_tag = html_options[:tag] || :div
--- a/lib/react/rails/component_mount.rb
+++ b/lib/react/rails/component_mount.rb
@@ -11,11 +11,14 @@ module React
       attr_accessor :output_buffer
       mattr_accessor :camelize_props_switch
 
+      def initialize
+        @cache_ids = []
+      end
+
       # {ControllerLifecycle} calls these hooks
       # You can use them in custom helper implementations
       def setup(controller)
         @controller = controller
-        @cache_ids = []
       end
 
       def teardown(controller)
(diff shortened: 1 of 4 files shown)


BookOfGreg:
Hi @HorizonShadow ,
First I'd like to say it's been a while since I was last able to help out here so I'd understand if you're not interested in helping this particular PR through anymore.
I do want to say this looks both valuable and useful to those using Turbolinks so I'd like to help.
I managed to correct the entire test suite for react-rails, basically it was a cascade of old dependencies, JS packages get made and deprecated every week it feels like so keeping up is a hell of a treadmill!
Now that that's fine you should be able to merge master back into your branch and see it cleanly.

If you're still getting gyp related issues, that's a common NodeJS issue that happens for a whole variety of reasons, rm -rf the node_modules (And the ones in `test/dummy_apps`) usually goes some way to helping.
The other is becuase of the complex support structure of this gem, bundler is stuck at 1.x while we support rails 4.2 still. You may have to `gem uninstall bundler` and `gem install bundler -v 1.17.3`

Hopefully this will get us going again. I pulled your branch in locally and their tests now pass on my machine!

BookOfGreg:
The one with tests #978 was merged into 2.6.0
This branch can now be removed. Thank you for contributing to the gem!

PR #961, opened 2019-01-30
Fix assert_react_component test helper for multiple components page

huacnlee:
Last  commit https://github.com/reactjs/react-rails/commit/71ced0c2ac5b86c9393a5476436bf25a61786983 have bug.

When the page have more than one React Component, `assert_react_component` can't catch the next items.

PR #934, opened 2018-09-15
Chore/update appraisals

BookOfGreg:
### Summary

Adds Rails 5.2 + Webpacker 3, 4.pre versions to appraisal, with Webpacker 4.pre set as allow fail.
Closes #918

PR #974, opened 2019-03-31
Fix/assert react component with multiple case

BookOfGreg:
### Summary

Extension to #961

This corrects the issue of webpacker 3 appearing to not compile when it was an execution-order dependent test. This was fixed by the `WebpackerHelpers.compile_if_missing` entry.

This includes extensions to test views in `{ name: @name, last_name: "Last #{@name}", info: { name: @name } }`  which completes the above linked PR.


BookOfGreg:
The only failures are due to Travis's version of bundler. Will correct in #934

PR #980, opened 2019-04-17
Polyfill loadable-stats

spajic:
Polyfill loadable-stats

### DIFF
--- a/lib/react/server_rendering/exec_js_renderer.rb
+++ b/lib/react/server_rendering/exec_js_renderer.rb
@@ -30,12 +30,12 @@ module React
       def global_wrapper
         @_global_wrapper ||= begin
           # Add loadable_stats.json
-          stats = File.read('public/packs/loadable-stats.json')
+          manifest = File.read('public/packs/manifest.json')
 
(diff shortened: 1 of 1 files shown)


lib/react/server_rendering/exec_js_renderer.rb
```diff
@@ -27,10 +27,18 @@ def before_render(component_name, props, prerender_options); ''; end
...
 
       # Handle Node.js & other ExecJS contexts
-      GLOBAL_WRAPPER = <<-JS
-        var global = global || this;
-        var self = self || this;
-      JS
+      def global_wrapper
+        @_global_wrapper ||= begin
+          # Add loadable_stats.json
+          manifest = File.read('public/packs/manifest.json')
```
BookOfGreg: There's no guarantee that the manifest would appear at that path. Could you make use of the Webpacker (Or sprockets) helper methods to pull in the relevant manifest? Also this may want to conditionally set the below var dependent on a config setting somewhere?

alkesh26:
@justin808 This can be closed.

PR #990, opened 2019-05-27
Typescript component generator

### LINKED ISSUE #965: Generate Typescript based component
```
rails g react:component HelloTypescript --ts
```

### Expected behavior
```
Running via Spring preloader in process 1234
      create  app/javascript/components/HelloTypescript.ts(x)
```

### Actual behavior
```
Running via Spring preloader in process 1234
      create  app/javascript/components/HelloTypescript.js
```

-------

When I have run `bundle exec rails webpacker:install:typescript` or when I specify `--ts` after the generate `react:component`, I'd like it to create a Typescript component instead of a JS one. (I see there is a `--coffee` flag too, so a `--ts` flag would be nice)

delkopiso:
## Summary

Allow the generation of typescript components by passing the `--ts` flag to the generator

resolves #965 

### Other information
Also handles all the special prop types like `instanceOf`, `oneOf`, and `oneOfType`:
- Treat instanceOf prop types as reference to predefined types.
- Treat oneOf prop types as union types, and define the union type.
- Treat oneOfType prop types as union of primitives and custom types, and define both custom type and union type

#### Example
```bash
$ bin/rails g react:component HelloWorld greeting:string \
  'food:oneOf{meat,cheese,vegetable}' \
  'cuisine:oneOfType{string,Food}' \
  owner:instanceOf{Person} --ts
      create  app/javascript/components/HelloWorld.tsx
```

```typescript
// app/javascript/components/HelloWorld.tsx
import * as React from "react"

type Food = 'meat' | 'cheese' | 'vegetable'
type Cuisine = string | Food

interface IHelloWorldProps {
  greeting: string;
  food: Food;
  cuisine: Cuisine;
  owner: Person;
}

interface IHelloWorldState {
}

class HelloWorld extends React.Component <IHelloWorldProps, IHelloWorldState> {
  render() {
    return (
      <React.Fragment>
        Greeting: {this.props.greeting}
        Food: {this.props.food}
        Cuisine: {this.props.cuisine}
        Owner: {this.props.owner}
      </React.Fragment>
    );
  }
}

export default HelloWorld
```

### DIFF
--- a/lib/generators/react/component_generator.rb
+++ b/lib/generators/react/component_generator.rb
@@ -55,6 +55,11 @@ module React
                    default: false,
                    desc: 'Output es6 class based component'
 
+      class_option :ts,
+                   type: :boolean,
+                   default: false,
+                   desc: 'Output tsx class based component'
+
       class_option :coffee,
                    type: :boolean,
                    default: false,
@@ -92,6 +97,8 @@ module React
       def create_component_file
         template_extension = if options[:coffee]
           'js.jsx.coffee'
+        elsif options[:ts]
+          'js.jsx.tsx'
         elsif options[:es6] || webpacker?
           'es6.jsx'
         else
@@ -101,7 +108,12 @@ module React
         # Prefer webpacker to sprockets:
         if webpacker?
           new_file_name = file_name.camelize
-          extension = options[:coffee] ? 'coffee' : 'js'
+          extension = if options[:coffee]
+            'coffee'
+          elsif options[:ts]
+            'tsx'
+          else
+            'js'
           target_dir = webpack_configuration.source_path
             .join('components')
             .relative_path_from(::Rails.root)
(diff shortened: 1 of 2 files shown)


lib/generators/react/component_generator.rb
```diff
@@ -146,23 +187,58 @@ def webpacker?
...
+          if options[:ts]
+            { :name => name, :type => ts_lookup(name, type, args), :union => union?(args) }
+          else
+            { :name => name, :type => lookup(type, args) }
+          end
+        end
+      end
+
+      def union?(args = '')
+        return args.to_s.gsub(/[{}]/, '').split(',').count > 1
```
BookOfGreg: I was wondering how Rails manages to parse out it's command line generators with curly braces such as `rails g migration add_thing_to_table thing:string{10}`
Turns out it's very similar to what you've done here already:
https://github.com/rails/rails/blob/1a5381ff0cf04af68a50bd04f265b9b8199e37b4/railties/lib/rails/generators/generated_attribute.rb#L45
But yours is a little easier to read 👍

PR #1017, opened 2019-08-20
Support SSR with local webpack server over https

porter77:
### Summary

We have run into a certificate verification error when server rendering locally with a webpack dev server running behind https. Adding `OpenSSL::SSL::VERIFY_NONE` fixes the issue. 

Not sure if this is the best fix, I'm willing to explore other alternatives if there is a better suggestion?


BookOfGreg:
That should only fail when your local machine does not trust the root of whichever cert you are trying to accept.
Add your self-signed cert to your computer's trust store and that should clear that up.

I'd accept a PR that allowed turning down the SSL mode with a default of verify but not one that will turn it off by default.

alkesh26:
@justin808, The PR is old. We can close this.

PR #1018, opened 2019-08-22
modified installation instructions for rails6

udaykadaboina:
### Summary

Modified installation instructions for Rails 6 🎉

### Other Information

Rails 6 now ships with webpacker as default JS compiler. See:  https://github.com/rails/rails/pull/33079

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,15 @@ gem 'react-rails'
 ```
 
 ##### 3) Now run the installers:
+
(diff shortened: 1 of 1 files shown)


BookOfGreg:
I'll look to merging once the issue with webpack-cli is sorted.
We might consider upgrading one of the appraisal gemfiles to rails 6 also.

PR #1020, opened 2019-09-03
Call handleUnmount before rendering a new page

jkbailey:
### Summary

UJS would mount non permanent react components multiple times when navigating away and then back to a page with one.

### Other Information

This was modified back in July in [this commit](https://github.com/reactjs/react-rails/commit/4dbcd6756dcd09c312abf95a73cbb6026f3f9a69#diff-caee293a420e16032c5cec4539620c8b). It appears as if it might have been modified unintentionally.

### DIFF
--- a/react_ujs/src/events/turbolinks.js
+++ b/react_ujs/src/events/turbolinks.js
@@ -2,11 +2,11 @@ module.exports = {
   // Turbolinks 5+ got rid of named events (?!)
   setup: function(ujs) {
   	ujs.handleEvent('turbolinks:load', ujs.handleMount);
-    ujs.handleEvent('turbolinks:before-render', ujs.handleMount);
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Going to merge this just now. Thanks for contributing. If you could throw in a test to ensure it's not toggled back accidentally in another PR I'd really appreciate it also!
Commented in linked issue why it's taking a while to go through, hope you will bear with me or can offer a hand to get this out.

jnstq:
@BookOfGreg Would be great if we could get a new release with this fix included with react_ujs v2.6.0 the callback componentWillUnmount isn't triggered.

BookOfGreg:
@jnstq Yep I'll try get a release out at the weekend. I hate releasing while the tests don't pass but I can't really help that my dependencies didn't follow semver.

jnstq:
@BookOfGreg Just checking the status of this 😄

PR #1027, opened 2019-09-26
Changed Let to Var

### LINKED ISSUE #1026: `let` breaks old browsers
Hi.  Is it really necessary to use let here?

https://github.com/reactjs/react-rails/blob/dadaa74ea33c80d0076f735c4bdfcb87c882aa6e/react_ujs/index.js#L106

It throws an error
`SyntaxError: Expected an identifier but found 'component' instead`

upd: Oh i see this duplicate #1016

edisonaugusthy:
Changed Let to Var to avoid breaking down in older browsers
Will fix  #1026

### DIFF
--- a/react_ujs/index.js
+++ b/react_ujs/index.js
@@ -103,7 +103,7 @@ var ReactRailsUJS = {
         }
         throw new Error(message + ". Make sure your component is available to render.")
       } else {
-        let component = this.components[cacheId];
+        var component = this.components[cacheId];
         if(component === undefined) {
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Pretty frustrating that webpack-cli is bust at the moment but I see that this will fix the `let` issue for now.
I want to look into ways to avoid accidentally using new ES6 features in the future, either by compiling with babel for sprockets users or by some linting/tests/warning.

PR #1042, opened 2019-11-07
📝 Adds reference to sample repo demonstrating HMR

edelgado:
### Summary

Adds a link to a sample Rails app that demonstrates HMR working with `react-rails` and adds a note of the present incompatibility of SSR and HMR.


BookOfGreg:
The SSR HMR incompatibility shouldn't be that hard to sort out I believe, I've just never sat down to sort it.
There are certainly other folks doing it:
https://github.com/mavajee/example__react-ssr-with-hmr
Anyway until I get that sorted thank you for contributing this.

Issue #264, opened 2015-05-14
Debugging Solutions for Prerender


blainekasten:
In my experience, when using `prerender: true`, if something breaks the server side rendering, the error messages are less than helpful.. My workflow has become anytime that happens to stop prerendering and use browser tools to diagnose where the issue is. The stack trace provided by the gem is no where near the help that comes from browser tools.

Have we thought of the idea of in the case of a prerender error to not throw an error, but instead disengage the prerendering. Maybe wrap the code that caused the server error in a `debugger;` call to help users know something is wrong?

maybe even include an alert message. So something like this is what I have in mind:
#### prerender catches error
- find error location
- Inject the following above the error 

``` js
alert('ReactRails encountered an error server side rendering, please inspect your issue with your browser dev tools');
debugger;
// error code is here.
```

thoughts?

mchristen:
In my experience this is a pain point as well. My hackish remedy was to monkeypatch React::Renderer.reset_combined_js! to write out the contents of @@combined_js to a file in tmp/

You can at least this way match up line numbers from the stack trace to actual javascript code, still not ideal but has helped me solve problems with server side rendering.

blainekasten:
yeah. I could also see something cool like that being printed to the ruby console. a good example of an error indicator is the babel compiler. When it encounters errors, it spits out a syntax highlighted output of the error line and the surrounding lines. That could also be a great solution.

rmosolgo:
:+1: Recently I'm using server-rendering more myself, and I've had the same workflow (first make it perfect, then `prerender: true`). I'd be happy to pitch in or try something out along the way!

barefootford:
@rmosolgo Before you do server pre-rendering, do you get real React errors/debugging? Right now whenever there is a React error in my app, the component doesn't render and fails silently. No errors on rails server or client side in chrome console.

rmosolgo:
huh, yeah I usually get normal React errors (or a `such-and-such is undefined` if I have a typo in the `react_component` helper). Happy to help debug if you care to share some code. I think I'm on React 0.13, not sure if anything changed since then.

hundsim:
I second @barefootford, whatever error in React I only get "SyntaxError: unknown:" and reference to javascript_include_tag 'application': 

![react_error](https://cloud.githubusercontent.com/assets/1907343/9439878/22b96a0e-4a6b-11e5-8066-861e6817d87c.png)

It would be so helpful to get normal React errors! :)

[...]

PR #1050, opened 2020-01-10
Expose all built-in options for `getConstructor`

hibachrach:
### Summary

This enables us to easily override `getConstructor` to not use global
fallback, avoiding the all-consuming `try...catch`.

### Other Information

Here's the context:

Quoting https://github.com/reactjs/react-rails/issues/264#issuecomment-552326663

> Regarding `Encountered error "#<ExecJS::ProgramError: Invariant
> Violation: Element type is invalid: ...`:
>
> I think one of the core issues is that [module lookup uses
> `try...catch`](https://github.com/reactjs/react-rails/blob/master/react_ujs/src/getConstructor/fromRequireContextWithGlobalFallback.js#L11-L23).
> While the errors are logged to the console shim, that typically doesn't
> help as a later error (such as the invariant violation) will lead to a
> fatal error (triggering a 500). If that could be refactored to be a bit
> more intentional based on environment (instead of just reacting based on
> exceptions, or at the very least, throwing if the caught exception isn't
> very specific)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 #### New Features
 
+- Expose alternative implementations for `ReactUJS.getConstructor` #1050
+
 #### Deprecation
 
 #### Bug Fixes
--- a/README.md
+++ b/README.md
@@ -406,12 +406,26 @@ delete window.Turbolinks;
 
 ### `getConstructor`
 
-Components are loaded with `ReactRailsUJS.getConstructor(className)`. This function has two built-in implementations:
+Components are loaded with `ReactRailsUJS.getConstructor(className)`. This function has two default implementations, depending on if you're using the asset pipeline or Webpacker:
 
-- On the asset pipeline, it looks up `className` in the global namespace.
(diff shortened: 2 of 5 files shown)


justin808:
@HarrisonB can you rebase this on master?

HarrisonB:
Sure! I'll hopefully get to it in the next week or two

HarrisonB:
@justin808 updated!

justin808:
@HarrisonB can you confirm that all unit tests run locally?

Currently, we don't have CI working. We need to move to GH Actions.

HarrisonB:
Having some trouble--See https://github.com/reactjs/react-rails/pull/1198#issuecomment-1229123819

justin808:
@hibachrach is this ready for merge?

hibachrach:
yes, though we should have it trigger CI to make sure it passes. is there a way to do that on your end?

justin808:
@hibachrach did you rebase on master where the github actions are defined?

hibachrach:
Tests are now passing!

alkesh26:
@justin808 The PR is good to go if we have no review comments.

justin808:
Thanks @hibachrach!

PR #1056, opened 2020-03-05
Changing the default path of components in README.md

windhamwong:
### Summary

The default path of components should be in app/assets/javascript/component.


BookOfGreg:
Perhaps this is the case if you're using sprockets but that section of the Readme is for Webpacker and it does place it in the specified location.

Thanks for contributing but I can't accept this PR.

PR #1057, opened 2020-03-12
Fix broken Hyperloop link in Readme

paulodeon:
### Summary

The project name has changed and the old url was an ad

### Other Information

None


BookOfGreg:
Thank you for the update.

PR #1060, opened 2020-03-27
Add nested component call info

BabsLabs:
### Summary

Added instructions for how to properly call nested components to the project read me.

### Other Information

This is a simple read me change that will help new gem users.


BookOfGreg:
Thanks for the contribution.

PR #1063, opened 2020-04-25
add render component extension point

edmond-c:
### Summary

Add extension point to delegate component rendering to an override-able function, similar to `getConstructor`.

This allows custom implementations that renders parent wrapper components, e.g. 
```js
function(renderFunctionName, component, node, props) {
  const contextObj = JSON.parse(node.getAttribute("data-context-object"));

  ReactDOM.render(
    <SomeReactContext.Provider value={contextObj}>
      {component}
    </SomeReactContext.Provider>,
    node,
  )
}
```

Built-in support for hot reloading 🔥 would also be added via this extension point. See this PR: 

### Other Information

Inline comments have been added to discusses options considered and their tradeoffs.

Also uncertain how to fix the Travis CI build error


react_ujs/index.js
```diff
@@ -6,6 +6,8 @@ var detectEvents = require("./src/events/detect")
 var constructorFromGlobal = require("./src/getConstructor/fromGlobal")
 var constructorFromRequireContextWithGlobalFallback = require("./src/getConstructor/fromRequireContextWithGlobalFallback")
 
+var renderWithReactDOM = require("./src/renderComponent/withReactDOM")
```
edmond-c, line 9: folder structure and import pattern is taking reference from `getConstructor` for consistency

react_ujs/index.js
```diff
@@ -80,6 +82,11 @@ var ReactRailsUJS = {
     return ReactDOMServer[renderFunction](element)
   },
 
+  // Render `component` using the specified `renderFunction` from `react-dom`.
+  // Override this function to render components in a custom way.
+  // function signature: ("hydrate" | "render", component, node, props)
+  renderComponent: renderWithReactDOM,
```
edmond-c, line 88: The comment and pattern here intends to mirror `getConstructor`. Let me know if

1. the function signature can be better represented
1. if the type of `renderFunctionName` (`"hydrate" | "render"`) should be made more generic, i.e. `string`
1. export the `renderWithReactDOM` function or assign it in a constant, e.g. `ReactRailsUJS.DEFAULT_RENDER_COMPONENT_FUNCTION`

Reference:
https://github.com/reactjs/react-rails/blob/d5da11129459cd75fd003c75319b1f7440c37322/react_ujs/index.js#L63-L66

react_ujs/index.js
```diff
@@ -112,9 +119,9 @@ var ReactRailsUJS = {
         }
 
         if (hydrate && typeof ReactDOM.hydrate === "function") {
-          component = ReactDOM.hydrate(component, node);
+          renderComponent("hydrate", component, node, props);
         } else {
-          component = ReactDOM.render(component, node);
+          renderComponent("render", component, node, props);
         }
```
edmond-c, line 125: Assignment to `component` is removed here. It appears unused, since the function returns immediately afterwards. In addition, the return value from `ReactDOM.render` has been marked as [deprecated](https://reactjs.org/docs/react-dom.html#render).

> `ReactDOM.render()` currently returns a reference to the root `ReactComponent` instance. However, using this return value is legacy and should be avoided because future versions of React may render components asynchronously in some cases. If you need a reference to the root `ReactComponent` instance, the preferred solution is to attach a [callback ref](https://reactjs.org/docs/more-about-refs.html#the-ref-callback-attribute) to the root element.

react_ujs/src/renderComponent/withReactDOM.js
```diff
@@ -0,0 +1,10 @@
+var ReactDOM = require("react-dom")
+
+// Render React component via ReactDOM, for example:
+//
+// - `renderComponent("hydrate", component, node, props)` -> `ReactDOM.hydrate(component, node);`
+// - `renderComponent("render", component, node, props)` -> `ReactDOM.render(component, node);`
+//
+module.exports = function(renderFunctionName, component, node) {
+  ReactDOM[renderFunctionName](component, node);
+};
```
edmond-c, line 10: The last argument `props` here is unused in this default render function. However, it is provided to support extensibility for custom implementations. The hot reload variation, for uses `props` to re-render with a freshly-reloaded component constructor.

Let me know if
1. function should declare the `props` argument even if it's unused (for self-documentation/ease of copy-and-pasting)
1. function signature should change to pass `constructor` instead of/in addition to `component`. The existing implementation of cached turbolinks components would need extra considerations if `component` is not passed to this function

Previous proposal referenced: https://github.com/reactjs/react-rails/pull/595#issuecomment-243638753

> - from
> ```js
> ReactDOM.render(React.createElement(constructor, props), node);
> ```
> - to
> ```js
> ReactRailsUJS.renderComponent(node, constructor, props)
> 
> // .... then ... 
> renderComponent: function(node, constructor, props) { 
>   ReactDOM.render(React.createElement(constructor, props), node);
> },
> ```

edmond-c:
see:

PR #1099, opened 2021-01-02
Fix Warning/Error Caused by URI.open instead of open

ksylvest:
### Summary

Using open via the following snippet:

    require 'open-uri'
    open('https//...')

In ruby 2 results in the following deprecation warning:

    warning: calling URI.open via Kernel#open is deprecated
    call URI.open directly or use URI#open

In ruby 3 results in the following exception:

    Errno::ENOENT (No such file or directory @ rb_sysopen - https://...)

The fixes by calling open through URI.


BookOfGreg:
Thanks!

PR #1100, opened 2021-01-31
Update test dependencies

wasabigeek:
### Summary

Attempt to fix failing test suite.


wasabigeek:
@BookOfGreg wondering if you have any pointers on how best to approach this? I was trying to start by getting `rails_5.2_no_sprockets_webpacker_3` to work, got the what I think works for package.json, at least for running on my mac locally, then installing on trusty (see [Dockerfile](https://github.com/reactjs/react-rails/pull/1100/files#diff-438419edcea4aadfbc621de3620e27f8cd0b3e644669c252e3b1eda083cfe985)). But now the gemfiles are failing, and I'll be trying to figure out the ruby installation.

Some questions:
- Do you think we could move off trusty?
- What is the minimal node version we should try to pin to?
- How would you handle certain OS specific gems e.g. I deleted gemfile.lock because `libv8 (7.3.492.27.1-x86_64-darwin-18)` was failing to install in the CI?

PR #1102, opened 2021-02-24
📝: Update README.md [ci skip]

jae-mon:
### Summary

Added a note to point 2 of the header 'Get Started With Webpacker'.
For Rails 6, the line `gem 'webpacker'` is not needed in the gemfile, as
it is already added by default.

### Other Information

Leaving in the line generates an error when bundle install is run:
"You cannot specify the same gem twice with different version requirements. 
You specified: webpacker (>= 0) and webpacker (~> 4.0)"

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ $ cd my-app
 gem 'webpacker'
 gem 'react-rails'
 ```
+Note: For Rails 6, You don't need to add `gem 'webpacker'` to your gemfile in step 2 above.
+Webpacker is the default javascript compiler for Rails 6, and is already added to your gemfile
+when you create a new app.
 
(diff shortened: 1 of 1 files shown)


BookOfGreg (approved):
Approved. I'd also welcome a PR that changes the default to assume webpack and have a note to add if less than Rails 6. I'd imagine there are more new Rails 6 projects today than 5 so.

jae-mon:
> Approved. I'd also welcome a PR that changes the default to assume webpack and have a note to add if less than Rails 6. I'd imagine there are more new Rails 6 projects today than 5 so.

Thank you for approving my PR. I will also work on the PR you suggested for changing the default to assume webpack.

PR #1107, opened 2021-03-26
fix(README): clearer instructions in steps 3 and 4

thoran:
It was confusing to have a section in Step 3 which made reference to Rails 5.x and then have comments about Rails < 5.x inside whilst the remainder was identical to the section for Rails 6.x.

Similarly, it was confusing to make reference to Step 4 inside Step 3 for no good reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,20 +75,18 @@ gem 'react-rails'
 
 ##### 3) Now run the installers:
 
-###### Rails 6.x:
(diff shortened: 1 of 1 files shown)


BookOfGreg:
Duplicate of https://github.com/reactjs/react-rails/pull/1136
Thanks for contributing but someone else beat you to it.

Issue #1028, opened 2019-09-26
Components not cleaned up with turbo links navigation


mmccall10:
### Steps to reproduce
Navigate with turbolinks to pages with render_component.

### Expected behavior
Components should be unmounted when navigating to different pages. 

### Actual behavior
Each page visit renders a new component while leaving the previous ones in place. With turbolinks cached pages subsequent page visits renders 2 new components from what I assume is the cached page, then the new page content. 

### System configuration
**Sprockets or Webpacker version**:  webpacker (4.0.7)
**React-Rails version**: react-rails (2.6.0)
**Rect_UJS version**: react_ujs@^2.6.0
**Rails version**: 6.0
**Ruby version**:  2.6.3

-------

It looks like a fix for this might have been submitted with PR #1020. That is almost a month old so maybe its a different issue?

Adding the following code to application.js seems to work as a temporary work around.
```js
document.addEventListener('turbolinks:before-render', () => {
    ReactRailsUJS.handleUnmount()
});
```
It appears the default way ReactRailsUjs is handling unmounting with turbolinks is the issue. 

Sample app that recreates the problem. 
https://github.com/mmccall10/turbolinks-react-rails-demo

marcelokanzaki:
I can confirm that. @mmccall10 thanks for the quick fix. Let's wait for the PR now...

jkbailey:
Thanks for mentioning my PR, also thanks for the quick fix for now. Not sure how to speed up the PR though.

BookOfGreg:
The problem I'm having with PRs at the moment is mostly around the general JS mess. Since Webpack split out Webpack-cli in a teeny patch, not following semantic versioning at all, the pipeline has been broken. Also Chrome stopped supporting the normal selenium API in the last release so I've been spending vast majority of my time just trying to get master built.

I recognize this is an issue for you and the linked PR is valid and correct, sorry this is taking so long. I'm really frustrated too.

mmccall10:
@BookOfGreg  thanks for response. No pressure! Also, thanks for confirming.

calleluks:
Hey @mmccall10 and @BookOfGreg, I'm probably missing something here so please excuse what might be a dumb question: 

When navigating with Turbolinks, why do we need to unmount the components?

Turbolinks uses [`cloneNode(true)`](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) when caching pages so we shouldn't risk leaving old event listeners around.

[...]

PR #1135, opened 2021-06-24
Don't unmount components on Turbolinks navigation

calleluks:
[As discussed in #1028](https://github.com/reactjs/react-rails/issues/1028#issuecomment-862475500), unmounting components on `turbolinks:before-render` prevents Turbolinks from restoring the scroll position when navigating back to pages with tall React components. This happens because Turbolinks tries to restore the scroll position before firing the `turbolinks:load` event. Since components were unmounted prior to this, the page might be much shorter than it will be after mounting components causing scroll position restoration to fail.

We've been running this in production for a week now and haven't noticed any issues caused by not unmounting components.

### DIFF
--- a/react_ujs/src/events/turbolinks.js
+++ b/react_ujs/src/events/turbolinks.js
@@ -2,11 +2,9 @@ module.exports = {
   // Turbolinks 5+ got rid of named events (?!)
   setup: function(ujs) {
   	ujs.handleEvent('turbolinks:load', ujs.handleMount);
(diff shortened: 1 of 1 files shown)


henrik:
@BookOfGreg Hi! Would it be possible to get a new react_ujs release incorporating this fix?

Or at least to get it into react_ujs/dist?

We'd love to use it via https://www.npmjs.com/package/react_ujs.

henrik:
@BookOfGreg Hi! Sorry to bother you – just wanted to check again if I could encourage either a new release or to get this change into dist?

PR #1148, opened 2021-11-14
Fix test suite

kylemellander:
### Summary

Currently, if you pull down the app to do any work to contribute, and you run the tests, there are many errors that will pop up preventing you from being able to run the test suite.

The intention of this PR is to get the tests in working order so that we can build more tests to make sure that support for newer features, like webpack 5, are supported.

The issues lies in 2 main places:

1. mimemagic has yanked 3.3.3 (up through 3.3.7), so this upgrades the version for the tests to 3.3.10.
2. The current version of webpack-dev-server in the dummy app does not support webpack 3, which it is built for.  This change lowers the version of `webpack-dev-server` to `2.11.5`, which is the last update that supports webpack 3.

### Other Information

This PR purposefully does not add more tests, but I intend to make another PR after this one that will do more to make sure that webpack 4 and 5 are testing building with the correct javascript libraries.  Currently, they are only built with the one in the dummy app which is statically locked to webpack 3, even though the tests refer to it running as webpack 4.

I think it could be beneficial to rename that build for now until we can get to making sure that it is properly tested.

In addition to the change, running the tests bumped the version of react_ujs in the package.json to the current version, so I thought it sane to commit it.  I can easily remove that if that should be handled by the travis run.

### DIFF
--- a/gemfiles/rails_5.2_no_sprockets.gemfile.lock
+++ b/gemfiles/rails_5.2_no_sprockets.gemfile.lock
@@ -127,7 +127,7 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (0.9.2)
-    mimemagic (0.3.3)
+    mimemagic (0.3.10)
     mini_mime (1.0.2)
     mini_portile2 (2.4.0)
     minitest (5.13.0)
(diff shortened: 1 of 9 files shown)


BookOfGreg:
Anyone working on stabilizing the tests here is :+1: in my books.

PR #1151, opened 2021-12-15
Support React 18's new client side rendering

kylemellander:
### Summary

React 18 is changing the way that client-side components are rendered and hydrated. See https://github.com/reactwg/react-18/discussions/5

In order to create a path forward to using React 18, this adds support for both methods of hydrating and rendering so that newer React features can be used while maintaining full backward compatibility.

### Other Information

I did not add any specific tests for this because there was no clear way to test for different versions of React.  This seems like it should be explored in separately.


BookOfGreg (approved):
It took me a second to understand why legacyReactRootLike was written the way it was, its clever. :+1:

PR #1186, opened 2022-06-29
BREAKING: Update to react-dom 18 client changes

ynagorny:
Update to import react-dom/client package instead of react-dom as described in:

https://reactjs.org/docs/react-dom.html

This takes care of this react 18 warning:

_You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client"._

### DIFF
--- a/lib/assets/javascripts/react_ujs.js
+++ b/lib/assets/javascripts/react_ujs.js
@@ -1,13 +1,13 @@
 (function webpackUniversalModuleDefinition(root, factory) {
 	if(typeof exports === 'object' && typeof module === 'object')
-		module.exports = factory(require("react-dom"), require("react"), require("react-dom/server"));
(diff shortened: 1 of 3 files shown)


justin808:
@ynagorny 

This seems to be a breaking change for those that don't upgrade to React 18. 

Please confirm.

Do you want to propose a solution that works with React 16+17 and React 18?

maedi:
Functions like `ReactDOM.hydrate` are still in the core library, only `createRoot` and `hydrateRoot` are in `react-dom/client` (see: https://github.com/facebook/react/blob/main/packages/react-dom/client.js).

I've created a PR (#1194) which uses the correct core/client APIs and attempts to provide backwards compatibility for React 16/17. Can someone test?

justin808:
Closed in favor of #1194.

PR #1269, opened 2023-05-01
Load react-dom correctly based on React version

### LINKED ISSUE #1258: Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".
### Steps to reproduce

1. `import ReactRailsUJS from 'react_ujs';` in your app
2. Make a call to `ReactRailsUJS.mountComponents(someElement);`
3. Visit a page that includes one or more React components rendered using react-rails (e.g. `<%= react_component 'Foo' %>`)
4. The following warning will appear in the JS console:
```
Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".
```
Trace:
image

With the breadcrumbs leading to the following culprit: https://github.com/reactjs/react-rails/blob/v2.7.0-rc.0/react_ujs/src/renderHelpers.js#L16

It appears that `renderHelpers.js` is importing the default `ReactDOM` from `react-dom`, but at the time that it accesses `ReactDOM.createRoot` the warning shows up.

### Expected behavior
ReactRailsUJS should be able to render components without triggering any warnings

### Actual behavior
Warning is triggered. This isn't a showstopper yet, but will be once React 19 is released, and creates a lot of noise in the JS console.

### System configuration
- **Shakapacker or Sprockets version**: we are using `vite_ruby 3.2.15`
- **React-Rails version**: `2.6.1` and `2.7.0.rc.0`
- **React_UJS version**: `2.6.2` and `2.7.0.rc.0`
- **Rails version**: `6.1.7.2`
- **Ruby version**: `3.2.0`

ahangarha:
### Summary

This PR fixes the warning related to loading `react-dom` in React 18.

### Pull Request checklist

- [x] ~Update documentation~
- [x] Update CHANGELOG file

### Other Information

Closes #1258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ _Please add entries here for your pull requests that are not yet released._
 
 #### Bug Fixes
 - Fix installation crash caused by absolute path for `source_entry_path` in default `config/webpacker.yml` coming from `shakapacker` version 6.x - #1216
+- Fix warning for loading `react-dom` in React 18 - #1269
 
 ## 2.6.2
 
--- a/react_ujs/src/reactDomClient.js
+++ b/react_ujs/src/reactDomClient.js
@@ -1,0 +1,24 @@
+import ReactDOM from "react-dom"
+
+const reactMajorVersion = ReactDOM.version?.split('.')[0] || 16
+
+// TODO: once we require React 18, we can remove this and inline everything guarded by it.
+const supportsRootApi = reactMajorVersion >= 18
+
+let reactDomClient = ReactDOM
+
+if (supportsRootApi) {
+  // This will never throw an exception, but it's the way to tell Webpack the dependency is optional
+  // https://github.com/webpack/webpack/issues/339#issuecomment-47739112
+  // Unfortunately, it only converts the error to a warning.
+  try {
+    // eslint-disable-next-line global-require,import/no-unresolved
+    reactDomClient = require('react-dom/client');
+  } catch (e) {
+    // We should never get here, but if we do, we'll just use the default ReactDOM
+    // and live with the warning.
+    reactDomClient = ReactDOM;
+  }
+}
+
+export default reactDomClient
--- a/react_ujs/src/renderHelpers.js
+++ b/react_ujs/src/renderHelpers.js
@@ -1,4 +1,4 @@
-import ReactDOM from "react-dom"
+import ReactDOM from "./reactDomClient"
 
 export function supportsHydration() {
   return typeof ReactDOM.hydrate === "function" || typeof ReactDOM.hydrateRoot === "function"


justin808 (approved):

gregorbg:
The code comment pointing to https://github.com/webpack/webpack/issues/339#issuecomment-47739112 specifies that:

> This will succeed on runtime and only print a warning while compiling (instead of an error).

This statement is correct, i.e. our frontend code compiles without error, but the warning is displayed as a transparent overlay when running webpacker dev server. It is also constantly printed to our logs. Is there a more elegant way than shutting webpacker up via config? Why is it trying to even import `react-dom/client` anyways when we are clearly referencing `"react-dom": "^16.14.0"` in our `package.json`?

ahangarha:
@gregorbg Do you mean you still get the warning message in v2.7?
I personally tested on two different setups and I didn't notice any warning.

gregorbg:
I am specifically saying that we get the error **only** on v2.7
Reverting to 2.6.1 makes the warning disappear.

We are using React and React-DOM v16.14.0 and our code is at https://github.com/thewca/worldcubeassociation.org. To reproduce, literally just spin up the Docker container with `docker-compose up`.

gregorbg:
Thanks for pointing me in the right direction! However, I can't help but wonder the following:

The webpacker error message points me to `WARNING in ./node_modules/react_ujs/react_ujs/src/reactDomClient.js 16:21-48`. Looking at the source code, this file contains the following check:

```js
import ReactDOM from "react-dom"

const reactMajorVersion = ReactDOM.version?.split('.')[0] || 16

// TODO: once we require React 18, we can remove this and inline everything guarded by it.
const supportsRootApi = reactMajorVersion >= 18

let reactDomClient = ReactDOM

if (supportsRootApi) {
  // This will never throw an exception, but it's the way to tell Webpack the dependency is optional
  // https://github.com/webpack/webpack/issues/339#issuecomment-47739112
  // Unfortunately, it only converts the error to a warning.
  try {
    // eslint-disable-next-line global-require,import/no-unresolved
    reactDomClient = require('react-dom/client');
  } catch (e) {
    // We should never get here, but if we do, we'll just use the default ReactDOM
    // and live with the warning.
    reactDomClient = ReactDOM;
  }
}

export default reactDomClient
```

where the relevant line is `reactDomClient = require('react-dom/client');`. The import is guarded by an if-check that basically reads "is the ReactDOM major version greater than or equal to 18?". In our case, we are using React 16 meaning this should return false, meaning this `if` should never be executed.

I can certainly implement the fix that you mentioned (and pointed us to in https://github.com/thewca/worldcubeassociation.org/issues/7904, thanks a lot! :smile:) but **why is this warning even being shown in the first place**?

ahangarha:
Thanks @gregorbg for your feedback.

tf:
This change appears to break SSR. See #1276.

PR #1291, opened 2023-07-20
Changes necessary for easy local testing

Judahmeek:
These changes rename the testing gemfiles, change some conditional logic on test/dummy/config/application, & remove references to action_mailer.

These changes enable devs to use `bundle exec ruby -Ilib:test:.:lib:test -e 'require "minitest/autorun"; require "relative-path-to-test";'` from the root directory to locally test individual test files.


ahangarha (approved):
This is cool 👍🏾

PR #1273, opened 2023-05-03
replace nulls in props from Rails with undefined

MaySoMusician:
### Summary

closes: discussion #1272 

I'm currently working a Rails+React project and eliminating `null`s from the type definitions in the frontend as possible, since my team members and I have been bothered about differentiation between `null` and `undefined` in terms of typing.

This PR adds an opt-in option to make react_ujs replace all `null`s in the props from Rails with `undefined`s, alongside the option system to allow to selectively enable each option. I've made the option default to `false` so that all existing projects won't break.

Read the [discussion](https://github.com/reactjs/react-rails/discussions/1272) for the more detailed background.

### Other Information

As soon as this gets merged, I'll also update [@types/react_ujs](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react_ujs) to add type definitions of `setOptions(option)`

### Pull Request checklist

- [ ] Add/update test to cover these changes
- [x] Update documentation
- [x] Update CHANGELOG file

PR #1293, opened 2023-07-27
Replace nulls in props from Rails with undefined

ahangarha:
Update/Rebase to master for #1273 PR.


Judahmeek (changes requested):
@ahangarha we need a test for this functionality

MaySoMusician:
I tried to set up the test environment with appraisal but couldn't get it. At least we have to upgrade appraisal to the latest because the currently-specified version of it doesn't correctly install the dependencies by `bundle exec appraisal install`

ahangarha:
> the currently-specified version of it doesn't correctly install the dependencies by bundle exec appraisal install

@MaySoMusician May you be more specific? Have you tried to work from this branch?

ahangarha:
With changes in [e77835e](https://github.com/reactjs/react-rails/pull/1293/commits/e77835e16386ba54f5276b326917d7c1fffe5ebd), the logic is implemented in Ruby side. The implementation may seem tricky and risky, but I think because of the nature of json, I could manage to match only `key":null` cases and ensure no other `null` strings would get replaced. 

I would like to see if any input can break this logic. I included a highly possible similar match in the tests to ensure it doesn't get replaced.

@MaySoMusician May you please check if this solution offers the exact result you are looking for?

lib/react/rails/component_mount.rb
```diff
@@ -67,6 +67,15 @@ def generate_html_options(name, options, props, prerender_options)
         html_options
       end
 
+      def props_to_json(props)
+        return props if props.is_a?(String)
+        return props.to_json unless Dummy::Application.config.react.null_to_undefined_props
+
+        # This regex matches key:value with null values while ensuing no string with similar
+        # pattern gets matched. It doesn't include null values in arrays.
+        props.to_json.gsub(/([^\\]":)null([,}\]])/, '\1undefined\2')
```
justin808, line 76: Regexps are great for many units tests to demonstrate the regexp is correct.

react_ujs/index.js
```diff
@@ -108,7 +118,8 @@ var ReactRailsUJS = {
       var className = node.getAttribute(ujs.CLASS_NAME_ATTR);
       var constructor = ujs.getConstructor(className);
       var propsJson = node.getAttribute(ujs.PROPS_ATTR);
-      var props = propsJson && JSON.parse(propsJson);
+      var props = propsJson && (ujs.options.replaceNull ? replaceNullWithUndefined(JSON.parse(propsJson))
```
justin808, line 121: why not a separate var for the parse result to avoid duplicating that bit?

justin808 (changes requested):

react_ujs/index.js
ahangarha: If the ruby solution is good, all js implementations will be removed.

ahangarha:
The commits for updating lock files are extracted into a PR #1297 to keep the PR focused

justin808:
Let's rebase on master.

lib/react/rails/component_mount.rb
```diff
@@ -56,7 +56,10 @@ def generate_html_options(name, options, props, prerender_options)
         unless prerender_options == :static
           html_options[:data].tap do |data|
             data[:react_class] = name
-            data[:react_props] = (props.is_a?(String) ? props : props.to_json)
+            data[:react_props] = props_to_json(
+              props,
+              null_to_undefined: Dummy::Application.config.react.null_to_undefined_props
```
lcmen, line 61: @justin808 should we use `Dummy::Application` here? `Dummy` is just for tests. I suspect it should be `Rails.configuration` / `Rails.application.config` (both are the same), no?

lcmen:
Hey! I believe this PR makes our controller specs to crash with a following error:

```
ActionView::Template::Error:
       #   uninitialized constant React::Rails::ComponentMount::Dummy
```

It comes from `Dummy::Application` reference in `lib/react/rails/component_mount.rb:61` line.

lib/react/rails/component_mount.rb
justin808: Yes. Agree.

@ahangarha can you get a PR in for this ASAP?

Or @lcmen if you can get us a PR, I'll merge ASAP. 

Extra credit if you provide a test that shows this bug.

ahangarha: I think the PR was not ready for merge. To me it was still a WIP with good progress.

I am on it

ahangarha:
@lcmen 
Thanks for reporting the issue.

MaySoMusician:
Hmm... I don't think Ruby can handle this problem due to the JSON spec. A JSON can't contain `undefined` value so that, in the current implementation, react_ujs will fail to [`mountComponents`](https://github.com/reactjs/react-rails/blob/910bb9952f27118ef9124abecbbfc1af51ec4327/react_ujs/index.js#L111) where it parse the JSON string from Rails.

My [very first post in the discussion](https://github.com/reactjs/react-rails/discussions/1272#discussion-5151389) has already pointed this out.
> react-rails passes the values from Rails to React via `data-react-props` HTML attribute with JSON, in the specification of which `undefined` is not defined (no pun intended). Due to this, in the part of converting them to JSON, we have no choice than converting Ruby's `nil`s to JSON's `null`s.

Or if you missed my post, you can recognize this limitation by testing the library in a real application (or the test suite, if available...)

lcmen:
The more I think about this, the more I agree with @justin808 response that this should not be a responsibility of the library. What I would suggest as an alternative, is a some kind of serializer callback function that could be customized via `config` option. It would simply take `props` and run custom function to serialize props. Something like this (pseudo  code):

```ruby
def props_to_json(props, options = { null_to_undefined: false })
  return props if props.is_a?(String)
  
  return Rails.configuration.react_props_serializer.call(props) if Rails.configuration.react_props_serializer.present?

  # rest of the code
end
```

lib/react/rails/component_mount.rb
chtzvt: Hey folks, 

Just wanted to chime in that this exact issue is breaking the 3.1.0 gem for me on a fresh Rails app. 

Following the installation steps outlined in the README produces:

```
> rails s
=> Booting Puma
=> Rails 7.0.7 application starting in development
=> Run `bin/rails server --help` for more startup options
Puma starting in single mode...
* Puma version: 5.6.7 (ruby 3.1.2-p20) ("Birdie's Version")
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 19068
* Listening on http://127.0.0.1:3000
* Listening on http://[::1]:3000
Use Ctrl-C to stop
Started GET "/" for 127.0.0.1 at 2023-08-20 20:15:25 -0400
Processing by HomeController#index as HTML
  Rendering layout layouts/application.html.erb
  Rendering home/index.html.erb within layouts/application
  Rendered home/index.html.erb within layouts/application (Duration: 2.8ms | Allocations: 4135)
  Rendered layout layouts/application.html.erb (Duration: 2.9ms | Allocations: 4247)
Completed 500 Internal Server Error in 8ms (ActiveRecord: 0.0ms | Allocations: 8205)



ActionView::Template::Error (uninitialized constant React::Rails::ComponentMount::Dummy

              null_to_undefined: Dummy::Application.config.react.null_to_undefined_props
                                 ^^^^^):
    2: <p>Find me in app/views/home/index.html.erb</p>
    3:
    4: <!-- erb: paste this in view -->
    5: <%= react_component("HelloWorld", { greeting: "Hello from react-rails." }) %>

app/views/home/index.html.erb:5
^C- Gracefully stopping, waiting for requests to finish
=== puma shutdown: 2023-08-20 20:21:16 -0400 ===
- Goodbye!
Exiting
```

ahangarha: Please upgrade to 3.1.1.

PR #1299, opened 2023-08-16
Make generator es6 and ts usable at same time

ippachi:
### Summary

Thank you for great integration rails and react. I use react-rails recently and I notice that I can't use `--ts` and `--es6` same time. What about making these two available at the same time?

### Other Information

This is actual generated file

`rails g react:component HelloWorld greeting:string --ts --es6`

```tsx
import * as React from "react"

interface IHelloWorldProps {
  greeting: string;
}

const HelloWorld = (props: IHelloWorldProps) => {
  return (
    <React.Fragment>
      Greeting: {props.greeting}
    </React.Fragment>
  )
}

export default HelloWorld
```

### Pull Request checklist
- [x] ~Update documentation~
- [x] Update CHANGELOG file

### DIFF
--- a/lib/generators/templates/component.es6.tsx
+++ b/lib/generators/templates/component.es6.tsx
@@ -1,0 +1,24 @@
+<%= file_header %>
+interface I<%= component_name %>Props {
+  <% if attributes.size > 0 -%>
+  <% attributes.each do |attribute| -%>
+  <% if attribute[:union] -%>
+  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:name].titleize %>;
+  <% else -%>
+  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:type] %>;
+  <% end -%>
+  <% end -%>
+  <% end -%>
+}
+
+const <%= component_name %> = (props: I<%= component_name %>Props) => {
+  return (
+    <React.Fragment>
+      <% attributes.each do |attribute| -%>
+      <%= attribute[:name].titleize %>: {props.<%= attribute[:name].camelize(:lower) %>}
+      <% end -%>
+    </React.Fragment>
+  )
+}
+
+<%= file_footer %>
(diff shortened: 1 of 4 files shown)


lib/generators/templates/component.es6.tsx
```diff
@@ -0,0 +1,24 @@
...
+<% if attributes.size > 0 -%>
+<% attributes.each do | attribute | -%>
+<% if attribute[:union] -%>
+  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:name].titleize %>;
+<% else -%>
+  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:type] %>;
+<% end -%>
+<% end -%>
+<% end -%>
+}
```
ahangarha, line 12: I think It would be good to use indentation here.

Judahmeek (changes requested):
@ippachi Please resolve @ahangarha's comment & also please rebase to address the Changelog conflict.

lib/generators/templates/component.es6.tsx
ippachi: @ahangarha 
Do you mean this?

```erb
interface I<%= component_name %>Props {
  <% if attributes.size > 0 -%>
  <% attributes.each do | attribute | -%>
  <% if attribute[:union] -%>
  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:name].titleize %>;
  <% else -%>
  <%= attribute[:name].camelize(:lower) %>: <%= attribute[:type] %>;
  <% end -%>
  <% end -%>
  <% end -%>
}
```

But All `<%` in template have no indentation.

ahangarha: I can say the indentations there are not ideal. Better to make a little improvement in each PR.

ippachi: @ahangarha 
OK. I have tried fixing indentation.

ahangarha: I think it can still be improved, but also the issue is beyond the changes in this particular PR.

So, I don't want to block the PR for this specific reason.

@Judahmeek If there is no other issue except this one, I think we are good to move forward.

justin808:
Thank you @ippachi for your contribution.

PR #1301, opened 2023-08-16
Update changelog for 3.1.1

ahangarha:


Judahmeek (changes requested):
Let's be more specific.

Just state that we removed the replace-null functionality due a severe logic error.

ahangarha:
@Judahmeek, I applied the requested change. You are right. The description should be specific enough to help the reader understand what has happened.

PR #1302, opened 2023-08-19
test: update fixture gemfile locks

G-Rath:
### Summary

As part of https://github.com/shakacode/shakapacker/pull/349 I have to update the lockfiles which would fail due to `mini_racer` so I had to regenerate them completely resulting in a failure because of a newer version of `selenium-webdriver` being used.

I figure ultimately this change is useful on its own because it'll impact anyone that needs to regenerate the lockfiles in future, so I might as well do a dedicated PR for it.

### Pull Request checklist

- [x] ~Add/update test to cover these changes~
- [x] ~Update documentation~
- [x] ~Update CHANGELOG file~

PR #1306, opened 2023-09-02
feat: support other js package managers

G-Rath:
### Summary

See https://github.com/shakacode/shakapacker/pull/349 and https://github.com/shakacode/shakapacker/issues/195

### Other Information

This is also passing CI in our [`rails-template`](https://github.com/ackama/rails-template/pull/466).

### Pull Request checklist

- [x] Update documentation
- [x] Update CHANGELOG file

### DIFF
--- a/gemfiles/shakapacker.gemfile
+++ b/gemfiles/shakapacker.gemfile
@@ -2,7 +2,6 @@
 
 source "http://rubygems.org"
 
-gem "package_json", github: "G-Rath/package_json", branch: "add-bun-support"
 gem "rails", "~> 7.0.x"
 gem "shakapacker", github: "G-Rath/shakapacker", branch: "use-package_json"
 
--- a/gemfiles/shakapacker.gemfile
+++ b/gemfiles/shakapacker.gemfile
@@ -3,6 +3,6 @@
 source "http://rubygems.org"
 
 gem "rails", "~> 7.0.x"
-gem "shakapacker", "7.2.0.rc.0"
+gem "shakapacker", "7.2.0"
 
 gemspec path: "../"
--- a/gemfiles/shakapacker.gemfile
+++ b/gemfiles/shakapacker.gemfile
@@ -3,6 +3,6 @@
 source "http://rubygems.org"
 
 gem "rails", "~> 7.0.x"
-gem "shakapacker", github: "G-Rath/shakapacker", branch: "use-package_json"
+gem "shakapacker", "7.2.0.rc.0"
 
 gemspec path: "../"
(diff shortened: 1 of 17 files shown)


gemfiles/shakapacker.gemfile
```diff
@@ -3,6 +3,6 @@
 source "http://rubygems.org"
 
 gem "rails", "~> 7.0.x"
-gem "shakapacker", "7.0.2"
+gem "shakapacker", github: "G-Rath/shakapacker", branch: "use-package_json"
```
justin808, line 6: @G-Rath we can update this after the release, and then this PR is good to go.

justin808:
@ahangarha @G-Rath can I merge once CI passes?

G-Rath:
@justin808 yup - ci is failing for unrelated reasons (new version of `rubocop-minitest`, which I think actually has a bug)

G-Rath:
Confirmed this is a bug in the Minitest plugin: https://github.com/rubocop/rubocop-minitest/pull/290

G-Rath:
@justin808 @ahangarha my bugfix for `rubocop-minitest` has been published so if you re-run CI the rubocop job should be passing now 🤞

I'm not so sure about the other failures - they're definitely not related to this change and I swear I even fixed them before but frankly Selenium has become a big blur recently with the recent ecosystem shift away from `webdrivers`; let's get Rubocop failing and maybe do a rerun of the other tests in case it's just flake and we can go from there.

G-Rath:
It looks like this is because of using Ruby v2.7, and might not have been helped by my inexperience with annotate - when I first started working on this I had issues with the locked version of `mini_racer` and updating resulted in an error due to a newer version of `selenium-webdriver`.

However it turns out the new version of `mini_racer` is not compatible with Ruby 2.7 - because I was using Ruby 3.x locally it was happy to install the newer versions and when running on 2.7 bundler apparently decides to just silently install the older versions (because `--frozen` isn't being set - I think that might be enabled by default in newer versions of bundler).

What's really weird is this doesn't seem to be happening consistently - #1302 passed CI, and running locally in some cases it passes and others it fails due to not being able to find Chrome (which could be because I'm on WSL and the older versions of `selenium-webdriver` can't handle that).

Overall this feels like a very big "chicken and egg" situation which is probably easiest addressed by just dropping support for Ruby 2.7... I'll try to do a bit more digging and see if I can unblock this another way though.

G-Rath:
This is passing as I've applied the fix from #1328 - that should be landed first; it seems like we don't actually need to update the lockfiles for this to be passing, but I still think it would be worth also landing #1326 once its been rebased to include updating the lockfiles properly.

G-Rath:
@justin808 this should be good to go!

justin808:
@G-Rath Let me know if you need a release ASAP.

PR #1285, opened 2023-07-06
Update code base based on sp7

ahangarha:
### Summary

This PR updates the code base based on Shakapacker 7.

### Pull Request checklist

- [x] Add/update test to cover these changes
- [x] Update documentation
- [x] Update CHANGELOG file

PR #1274, opened 2023-05-14
SSR Fixes and better tests for v3.0.0

Judahmeek:
### Summary

_Remove this paragraph and provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together._

### Other Information

_Remove this paragraph and mention any other important and relevant information such as benchmarks._

### Pull Request checklist
_Remove this line after checking all the items here. If the item is not applicable to the PR, both check it out and wrap it by `~`._

- [ ] Add/update test to cover these changes
- [ ] Update documentation
- [ ] Update CHANGELOG file

PR #1317, opened 2023-10-18
Update README.md

okuramasafumi:
### Summary

https://github.com/reactjs/react-rails/blob/master/CHANGELOG.md

> If using Webpacker/Shakapacker, requires upgrading to Shakapacker v7 #1274 and #1285

It looks like react-rails v3 supports Shakapacker v7 only.

### Pull Request checklist

- [x] Update documentation
- [x] Update CHANGELOG file


ahangarha:
@okuramasafumi Thanks for the fix.

PR #1321, opened 2023-12-07
ci: run `rubocop` on pull requests

G-Rath:
### Summary

Currently rubocop failures are not properly coming through on pull requests because the workflow only runs on push - this is why e.g. #1320 was able to slip through

### Pull Request checklist

- [x] ~Update documentation~
- [x] ~Update CHANGELOG file~

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,6 +1,6 @@
 name: Rubocop
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   rubocop:


G-Rath:
@Judahmeek yup you're right and that would be my preference but my change matches the other workflow - I'm happy to do a follow up PR changing both workflows to run only on pushes to `master`

Judahmeek:
@G-Rath Fair enough. Although I would have just combined #1320, #1321, and said follow-up, I understand & respect your reasons for not doing so.

Btw, how did you notice my merge of #1320 so quickly?

G-Rath:
Follow up PR is here: #1322

> Btw, how did you notice my merge of https://github.com/reactjs/react-rails/pull/1320 so quickly?

I got a notification in my inbox and just... actioned it 😅

PR #1326, opened 2024-01-01
Fix test failures

ahangarha:
### Summary

This PR fixes some of the recent test failures.

### Other Information

The change is based on the discussions in https://github.com/SeleniumHQ/selenium/issues/13112.

### Pull Request checklist
- [x] ~Update documentation~
- [x] ~Update CHANGELOG file~

### DIFF
--- a/test/react/rails/react_rails_ujs_test.rb
+++ b/test/react/rails/react_rails_ujs_test.rb
@@ -10,14 +10,14 @@ class ReactRailsUJSTest < ActionDispatch::IntegrationTest
     def assert_greeting(page, greeting)
       assert page.has_content?(greeting), <<~MSG
         #{page.body}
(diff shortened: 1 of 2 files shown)


G-Rath:
@ahangarha next thing to try would be to use `headless=new`, based on https://github.com/SeleniumHQ/selenium/issues/13112

G-Rath:
See https://github.com/reactjs/react-rails/pull/1328#issuecomment-1874412445 - this looks like the fix for after updating `selenium-webdriver` but it still means the lockfiles are actually incorrect for Ruby 2.7, so I think we should land #1328 first then land this after a rebase + redoing the general update from #1302

ahangarha:
If the issue is resolved (it seems it is) I close this PR.

PR #1327, opened 2024-01-01
chore: switch default branch to `main`

G-Rath:
### Summary

This is the canonical default branch these days - might as well get the switch over with; once this is merged ideally the `master` branch should be renamed to `main` using GitHub's UI so that workflow runs and other such logs are retained.

### Pull Request checklist

- [x] Update documentation
- [x] ~Update CHANGELOG file~

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -3,7 +3,7 @@ name: Rubocop
 on:
   push:
     branches:
-      - 'master'
+      - 'main'
   pull_request:
 
 jobs:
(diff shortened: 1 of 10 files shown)


justin808 (approved):

tnir:
@justin808 This code broke the link from `/docs/get-started.md` to `/VERSIONS.md` as the default branch is still `master`. Any idea on changing the default branch from `master` to `main`?

https://github.com/reactjs/react-rails/blob/6e811b6bc3634112a269821ee2f5b512167d39ae/docs/get-started.md#L338

tnir:
See #1349

PR #1305, opened 2023-08-31
bug: Prevent roots from being re-created when using React 18

diogobeda:
### Summary

When dynamically adding components to a page by using, for example, Turbo streams and either manually calling `ReactRailsUJS.mountComponents` or passing `ReactRailsUJS.handleMount` to a Turbo event, there's a bug where the react roots are re-created on a node instead of being re-used. That resets the local state of components that were already mounted and may also cause visual glitches with the component flashing.

### Other Information

I made a [reproduction repo](https://github.com/diogobeda/react-rails-18-root-bug) to showcase the bug happening and as a test case for the fix. Feel free to try it out for yourselves.

Here's a gif from before the fix:

CleanShot 2023-08-31 at 10 37 40

And one from after the fix:

CleanShot 2023-08-31 at 10 40 58

Please let me know if there's anything else I should add to this PR 🙏 

### Pull Request checklist
_Remove this line after checking all the items here. If the item is not applicable to the PR, both check it out and wrap it by `~`._

- [x] Add/update test to cover these changes
- [ ] Update documentation
- [x] Update CHANGELOG file

PR #1328, opened 2024-01-01
ci: revert lockfile updates and don't allow lockfile updates when installing

G-Rath:
### Summary

This reverts #1302 as updates to versions that are not compatible with Ruby v2.7 and ends up causing more harm - I've also updated `bundle install` to error if the lockfile would be changed when installing in CI to prevent this from happening in future.

This is to address the immediate issue of CI just being broken right now - I expect that this will still cause problems with #1306 so will investigate further once this is landed.

### Pull Request checklist

- [x] ~Update documentation~
- [x] ~Update CHANGELOG file~


ahangarha:
@Judahmeek, shouldn't we update the `react-rails` version in these lock files?

G-Rath:
ugh right a release was done after this update - that's annoying since this is technically a revert; I've figured out how to resolve the problem with just `mini_racer` though so have pushed that up which also updates the lockfiles to be generally correct without upgrading other versions (i.e so the `react-rails` version should now be correct).

In terms of steps, I think we should:
  1. merge this PR, fixing the current CI and local actions (since that's broken currently due to `mini_racer`)
  2. redo #1302 with the `headless=new` fix, so that we're up to date (since we'll eventually end up updating `selenium-webdriver` which will require this, so might as well get it done) - #1326 could be used for that
  3. rebase #1305, #1306, and #1327

PR #1335, opened 2024-04-12
Prevent root from being recreated on react 18 (recreation of #1305)

Judahmeek:
This is a recreation of #1305 as Diogo didn't give us permission to update his fork.

- [ ] CI Passes
- [ ] Update example app, and reference it in docs. Maybe remove doc for example app is https://github.com/BookOfGreg/react-rails-example-app b/c that's outdated or update it.

PR #1336, opened 2024-04-13
Ci Fixes

Judahmeek:
This PR addresses linting and some other CI issues in preparation for #1335

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -89,7 +89,7 @@ jobs:
       with:
         persist-credentials: false
     - uses: actions/setup-node@v3
-    - run: npm -g install yalc ${{ matrix.js_package_manager.installer }}
+    - run: sudo npm -g install yalc ${{ matrix.js_package_manager.installer }}
     - run: yalc publish
     - name: Save root node_modules to cache
       uses: actions/cache@v3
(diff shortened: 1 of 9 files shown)


.github/workflows/ruby.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       with:
         persist-credentials: false
     - uses: actions/setup-node@v3
-    - run: npm -g install yalc ${{ matrix.js_package_manager.installer }}
+    - run: sudo npm -g install yalc ${{ matrix.js_package_manager.installer }}
```
G-Rath, line 92: Seems weird that we have to use `sudo` but not that concerned since this in a temp CI runner 😁

(I figure it might be related to supporting old versions of Node and what not)

G-Rath (approved):

Gemfile
```diff
@@ -4,3 +4,6 @@ source "http://rubygems.org"
 
 gemspec
 # This is an optional dev-dependency, required whenever sprockets is required
+gem "rubocop"
```
justin808, line 7: why not in the gemspec as a development dependency?

.github/workflows/ruby.yml
Judahmeek: The `sudo` is to resolve a `EACCES: permission denied` error that I was getting.

The other option I saw in my search was `--unsafe-perm`, but that's apparently obsolete.

Gemfile
Judahmeek: Because CI doesn't install development dependencies from the gemspec.

Do you want me to create a `Gemfile.development_dependencies` file like react_on_rails has?

PR #1367, opened 2025-12-06
fix: support `connection_pool` v3

### LINKED ISSUE #1366: Incompatible with v3 of `connection_pool`
v3 changed `ConnectionPool` to [take keyword args rather than positional args](https://github.com/mperham/connection_pool/blob/main/Changes.md#300), which breaks

https://github.com/reactjs/react-rails/blob/28ee68955f00261aecc07061ea6ea1dfc0831e06/lib/react/server_rendering.rb#L19

I've not dug deeper into this yet so that might not be the only change that is needed - I'd consider doing a new version locking `connection_pool` to `<= 2` for the time being

G-Rath:
### Summary

`connection_pool` v3+ requires keyword arguments instead of positional args, but also drops support for Ruby 3.2 so this has us check the version being used and do the right thing accordingly

### Other Information

Resolves #1366

### Pull Request checklist

- [x] ~Update documentation~
- [x] Update CHANGELOG file

### DIFF
--- a/lib/react/server_rendering.rb
+++ b/lib/react/server_rendering.rb
@@ -16,13 +16,7 @@ module React
     # @return [void]
     def self.reset_pool
       options = { size: pool_size, timeout: pool_timeout }
-
-      # ConnectionPool v3+ uses keyword arguments rather than positional
-      @pool = if ConnectionPool::VERSION[0] <= "2"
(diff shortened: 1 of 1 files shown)


izn:
Any updates on when this can be merged?

kapso:
thanks @G-Rath 

@justin808 any plans for doing a release? thanks

